PR #1, opened 2023-06-07
feat: propose stark-felt api

xJonathanLEI:
A very preliminary API surface for the `stark-felt` crate. It's largely modeled after `starknet-ff` but also modified based on my survey of other felt impls. Obviously this would need feedback from the prominent felt users.

Once we can agree on an API surface, IMO the best way to test its fitness would be for felt dependants to actually start trying to switch to this crate to see if they can get it to compile. Obviously it would just panic at runtime but this ensures our API is good enough.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starknet-io/types-rs/1)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,3 +31,22 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
+
+  no-std-build:
+    name: Test no_std support
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install latest nightly
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          override: true
+          components: rustfmt, clippy
+          target: thumbv6m-none-eabi
+
+      - name: Build stark-felt
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --package stark-felt --target thumbv6m-none-eabi --no-default-features
--- a/crates/stark-felt/Cargo.toml
+++ b/crates/stark-felt/Cargo.toml
@@ -12,4 +12,9 @@ readme = "README.md"
 
 [dependencies]
 bitvec = { version = "1.0.1", default-features = false }
-serde = { version = "1.0.163", default-features = false }
+serde = { version = "1.0.163", optional = true, default-features = false }
+
+[features]
+default = ["std", "serde"]
+std = []
+alloc = ["serde?/alloc"]
--- a/crates/stark-felt/Cargo.toml
+++ b/crates/stark-felt/Cargo.toml
@@ -11,3 +11,5 @@ description = "Field element type for Cairo."
 readme = "README.md"
 
 [dependencies]
+bitvec = { version = "1.0.1", default-features = false }
+serde = { version = "1.0.163", default-features = false }
--- a/crates/stark-felt/src/lib.rs
+++ b/crates/stark-felt/src/lib.rs
@@ -1,4 +1,12 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
 use bitvec::array::BitArray;
+
+#[cfg(target_pointer_width = "64")]
+pub type BitArrayStore = [u64; 4];
+
+#[cfg(not(target_pointer_width = "64"))]
+pub type BitArrayStore = [u32; 8];
 
 /// Definition of the Field Element type.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
@@ -53,12 +61,12 @@ impl Felt {
     }
 
     /// Converts to big-endian bit representation.
-    pub const fn to_bits_be(&self) -> BitArray<[u64; 4]> {
+    pub const fn to_bits_be(&self) -> BitArray<BitArrayStore> {
         todo!()
     }
 
     /// Converts to little-endian bit representation.
-    pub const fn to_bits_le(&self) -> BitArray<[u64; 4]> {
+    pub const fn to_bits_le(&self) -> BitArray<BitArrayStore> {
         todo!()
     }
 
@@ -342,7 +350,7 @@ mod arithmetic {
     }
 }
 
-// TODO: gate this behind a `serde` feature
+#[cfg(feature = "serde")]
 mod serde {
     use ::serde::{Deserialize, Serialize};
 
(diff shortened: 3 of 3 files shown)


Mirko-von-Leipzig:
There is (sometimes) a desire to know what the inside of the `Felt` looks like. For example, is `Felt::from_be_bytes` or `Felt::from_le_bytes` a no-op? This would inform decisions on how to store this type.

I realise this "leaks" the internal implementation in a way. I suppose if this is desired the user can simply check the source code himself? Or alternatively, one adds a `Felt::from_inner` or `Felt::from_raw` which then becomes a breaking change if the inner type is changed?

I'm unsure how much of a concern this should be tbh.

xJonathanLEI (commented):
> *[Reviewable](https://reviewable.io/reviews/starknet-io/types-rs/1)* status: 1 of 2 files reviewed, 8 unresolved discussions (waiting on @abdelhamidbakhta, @dan-starkware, and @spapinistarkware)

___
*[`crates/stark-felt/src/lib.rs` line 36 at r1](https://reviewable.io/reviews/starknet-io/types-rs/1#-NXLDAErEs47x3YUhTXY:-NYaVtAE7mK9BNJP4G7i:bss4a2b) ([raw file](https://github.com/starknet-io/types-rs/blob/be522bd08aa9671f634582aa5cb62fc123cd7e25/crates/stark-felt/src/lib.rs#L36)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Remove
</blockquote></details>

Removed.
___
*[`crates/stark-felt/src/lib.rs` line 51 at r1](https://reviewable.io/reviews/starknet-io/types-rs/1#-NXLDC5N5nN0sN4CrOM3:-NYaVvps-kvq3S-rvBHj:bss4a2b) ([raw file](https://github.com/starknet-io/types-rs/blob/be522bd08aa9671f634582aa5cb62fc123cd7e25/crates/stark-felt/src/lib.rs#L51)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Remove
</blockquote></details>

Removed.
___
*[`crates/stark-felt/src/lib.rs` line 66 at r1](https://reviewable.io/reviews/starknet-io/types-rs/1#-NXLDHzX4gP1Gptb41uE:-NYaW6DK4C8i7zuCCbz8:bofcu4o) ([raw file](https://github.com/starknet-io/types-rs/blob/be522bd08aa9671f634582aa5cb62fc123cd7e25/crates/stark-felt/src/lib.rs#L66)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

What is this? who is using this API?
</blockquote></details>

`pathfinder` (and also `starknet-crypto` since it steals code from `pathfinder`):

https://github.com/eqlabs/pathfinder/blob/eaab33692dc4057b0855a62188b8d5c6909abb2d/crates/stark_curve/src/field.rs#L55
___
*[`crates/stark-felt/src/lib.rs` line 95 at r1](https://reviewable.io/reviews/starknet-io/types-rs/1#-NXLDUps9g4vWiJ6rzyh:-NYaWa2sBVVhg_hTJG3e:b-fsijxj) ([raw file](https://github.com/starknet-io/types-rs/blob/be522bd08aa9671f634582aa5cb62fc123cd7e25/crates/stark-felt/src/lib.rs#L95)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Mention that there may be 2 roots for each square, and we will return the lower one.
</blockquote></details>

Comment updated.
___
*[`crates/stark-felt/src/lib.rs` line 106 at r1](https://reviewable.io/reviews/starknet-io/types-rs/1#-NXLDblQC6ofaK2f1J5M:-NYaWl0-9zrAEkEtfvxj:b-ahh9wq) ([raw file](https://github.com/starknet-io/types-rs/blob/be522bd08aa9671f634582aa5cb62fc123cd7e25/crates/stark-felt/src/lib.rs#L106)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

The exponent is not a field element. Since `x**p=x`, the exponent lives in `Z mod (p-1)`. Maybe just get something like u128? Do you know where this is used?
</blockquote></details>

I'm not sure where this is used. I added this because [LambdaWorks](https://github.com/lambdaclass/lambdaworks) has it.

Updated the param type anyways.
___
*[`crates/stark-felt/src/lib.rs` line 110 at r1](https://reviewable.io/reviews/starknet-io/types-rs/1#-NXLDyLFEmPs8n6OkXV1:-NYaXdFD3RqQbmuxVBEO:b-1xecbg) ([raw file](https://github.com/starknet-io/types-rs/blob/be522bd08aa9671f634582aa5cb62fc123cd7e25/crates/stark-felt/src/lib.rs#L110)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

What does this mean for field elements? It's not well defined. Is this actually used anywhere?
</blockquote></details>

As you know I'm not a cryptographer so I'm not exactly sure about this, but it's used in `starknet-crypto`:

https://github.com/xJonathanLEI/starknet-rs/blob/ed14f483bb5d5171e6775960a24ee07018f3ff7c/starknet-crypto/src/ecdsa.rs#L88-L118
___
*[`crates/stark-felt/src/lib.rs` line 116 at r1](https://reviewable.io/reviews/starknet-io/types-rs/1#-NXLEDYwCWcrszzYa5nK:-NYaXpbpEWZ-XYed22r6:bkjk06y) ([raw file](https://github.com/starknet-io/types-rs/blob/be522bd08aa9671f634582aa5cb62fc123cd7e25/crates/stark-felt/src/lib.rs#L116)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Same
</blockquote></details>

Same as above: used in `starknet-crypto`.
___
*[`crates/stark-felt/src/lib.rs` line 342 at r1](https://reviewable.io/reviews/starknet-io/types-rs/1#-NXLENl22VkcqoEXpuCG:-NYaXCLqDcckTvOfA8Eo:bd323px) ([raw file](https://github.com/starknet-io/types-rs/blob/be522bd08aa9671f634582aa5cb62fc123cd7e25/crates/stark-felt/src/lib.rs#L342)):*
<details><summary><i>Previously, spapinistarkware (Shahar Papini) wrote…</i></summary><blockquote>

Bit operations are not defined well on felt. in particular, (P-1) | 1 = ?
Do we need these?
</blockquote></details>

Agreed. I was debating whether to add these. Even if we separate into 2 types, these would still be ambiguous for the non-computational type.

Removed.

abdelhamidbakhta (approved):
Lgtm

PR #2, opened 2023-07-05
feat: Use `LambdaWorks`' `FieldElement` in `stark-felt` crate

fmoletta:
This PR implements the current api in the stark-felt crate using the [LambdaWorks](https://github.com/lambdaclass/lambdaworks) library

### DIFF
--- a/crates/stark-felt/src/arbitrary.rs
+++ b/crates/stark-felt/src/arbitrary.rs
@@ -1,0 +1,48 @@
+use lambdaworks_math::{field::element::FieldElement, unsigned_integer::element::UnsignedInteger};
+use proptest::prelude::*;
+
+use crate::Felt;
+const FIELD_HIGH: u128 = (1 << 123) + (17 << 64); // this is equal to 10633823966279327296825105735305134080
+const FIELD_LOW: u128 = 1;
+
+/// Returns a [`Strategy`] that generates any valid Felt
+fn any_felt() -> impl Strategy<Value = Felt> {
+    (0..=FIELD_HIGH)
+        // turn range into `impl Strategy`
+        .prop_map(|x| x)
+        // choose second 128-bit limb capped by first one
+        .prop_flat_map(|high| {
+            let low = if high == FIELD_HIGH {
+                (0..FIELD_LOW).prop_map(|x| x).sboxed()
+            } else {
+                any::<u128>().sboxed()
+            };
+            (Just(high), low)
+        })
+        // turn (u128, u128) into limbs array and then into Felt
+        .prop_map(|(high, low)| {
+            let limbs = [
+                (high >> 64) as u64,
+                (high & ((1 << 64) - 1)) as u64,
+                (low >> 64) as u64,
+                (low & ((1 << 64) - 1)) as u64,
+            ];
+            FieldElement::new(UnsignedInteger::from_limbs(limbs))
+        })
+        .prop_map(Felt)
+}
+
+/// Returns a [`Strategy`] that generates any nonzero Felt
+pub fn nonzero_felt() -> impl Strategy<Value = Felt> {
+    any_felt().prop_filter("is zero", |x| !x.is_zero())
+}
+
+impl Arbitrary for Felt {
+    type Parameters = ();
+
+    fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
+        any_felt().sboxed()
+    }
+
+    type Strategy = SBoxedStrategy<Self>;
+}
--- a/crates/stark-felt/src/arbitrary.rs
+++ b/crates/stark-felt/src/arbitrary.rs
@@ -6,6 +6,7 @@ const FIELD_HIGH: u128 = (1 << 123) + (17 << 64); // this is equal to 1063382396
 const FIELD_LOW: u128 = 1;
 
 /// Returns a [`Strategy`] that generates any valid Felt
+/// This is used to generate input values for proptests
 fn any_felt() -> impl Strategy<Value = Felt> {
     (0..=FIELD_HIGH)
         // turn range into `impl Strategy`
@@ -33,6 +34,7 @@ fn any_felt() -> impl Strategy<Value = Felt> {
 }
 
 /// Returns a [`Strategy`] that generates any nonzero Felt
+/// This is used to generate input values for proptests
 pub fn nonzero_felt() -> impl Strategy<Value = Felt> {
     any_felt().prop_filter("is zero", |x| !x.is_zero())
 }
--- a/crates/stark-felt/src/lib.rs
+++ b/crates/stark-felt/src/lib.rs
@@ -2,18 +2,29 @@
 
 use bitvec::array::BitArray;
 
+#[cfg(test)]
+mod arbitrary;
+
 #[cfg(target_pointer_width = "64")]
 pub type BitArrayStore = [u64; 4];
 
 #[cfg(not(target_pointer_width = "64"))]
 pub type BitArrayStore = [u32; 8];
 
+use lambdaworks_math::{
+    field::{
+        element::FieldElement, fields::fft_friendly::stark_252_prime_field::Stark252PrimeField,
+    },
+    traits::ByteConversion,
+    unsigned_integer::element::UnsignedInteger,
+};
+
 /// Definition of the Field Element type.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-pub struct Felt {}
+pub struct Felt(FieldElement<Stark252PrimeField>);
 
 /// A non-zero [Felt].
-pub struct NonZeroFelt {}
+pub struct NonZeroFelt(FieldElement<Stark252PrimeField>);
 
 #[derive(Debug)]
 pub struct FeltIsZeroError;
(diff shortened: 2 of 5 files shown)


crates/stark-felt/src/arbitrary.rs
```diff
@@ -0,0 +1,48 @@
+use lambdaworks_math::{field::element::FieldElement, unsigned_integer::element::UnsignedInteger};
+use proptest::prelude::*;
+
+use crate::Felt;
+const FIELD_HIGH: u128 = (1 << 123) + (17 << 64); // this is equal to 10633823966279327296825105735305134080
+const FIELD_LOW: u128 = 1;
+
+/// Returns a [`Strategy`] that generates any valid Felt
```
LucasLvy, line 8: Can you elaborate more on when we should use this function ? more generally when do we need a `Strategy`.

crates/stark-felt/src/arbitrary.rs
```diff
@@ -0,0 +1,48 @@
...
+                (high & ((1 << 64) - 1)) as u64,
+                (low >> 64) as u64,
+                (low & ((1 << 64) - 1)) as u64,
+            ];
+            FieldElement::new(UnsignedInteger::from_limbs(limbs))
+        })
+        .prop_map(Felt)
+}
+
+/// Returns a [`Strategy`] that generates any nonzero Felt
```
LucasLvy, line 35: Same as above

crates/stark-felt/src/lib.rs
```diff
@@ -26,140 +37,193 @@ pub struct FromBytesError;
...
     /// Maximum value of [Felt]. Equals to 2^251 + 17 * 2^192.
-    pub const MAX: Self = Self {};
+    pub const MAX: Self = Self(FieldElement::<Stark252PrimeField>::const_from_raw(
+        UnsignedInteger::from_limbs([544, 0, 0, 32]),
+    ));
 
     /// Creates a new [Felt] from its big-endian representation in a [u8] slice.
-    pub const fn from_bytes_be(_bytes: &[u8]) -> Result<Self, FromBytesError> {
-        todo!()
+    pub fn from_bytes_be(bytes: &[u8]) -> Result<Self, FromBytesError> {
```
LucasLvy, line 59: can you add in the rust doc how fast is this function compared to the other ways of creating a felt.

crates/stark-felt/src/lib.rs
```diff
@@ -26,140 +37,193 @@ pub struct FromBytesError;
...
+        FieldElement::from_bytes_be(bytes)
+            .map(Self)
+            .map_err(|_| FromBytesError)
     }
 
     /// Creates a new [Felt] from its little-endian representation in a [u8] slice.
-    pub const fn from_bytes_le(_bytes: &[u8]) -> Result<Self, FromBytesError> {
-        todo!()
+    pub fn from_bytes_le(bytes: &[u8]) -> Result<Self, FromBytesError> {
```
LucasLvy, line 66: can you add in the rust doc how fast is this function compared to the other ways of creating a felt.

crates/stark-felt/src/lib.rs
```diff
@@ -26,140 +37,193 @@ pub struct FromBytesError;
...
+        FieldElement::from_bytes_le(bytes)
+            .map(Self)
+            .map_err(|_| FromBytesError)
     }
 
     /// Converts to big-endian byte representation in a [u8] array.
-    pub const fn to_bytes_be(&self) -> [u8; 32] {
-        todo!()
+    pub fn to_bytes_be(&self) -> [u8; 32] {
```
LucasLvy, line 73: can you add in the rust doc how fast is this function

crates/stark-felt/src/lib.rs
```diff
@@ -26,140 +37,193 @@ pub struct FromBytesError;
...
+        self.0.to_bytes_be()
     }
 
     /// Converts to little-endian byte representation in a [u8] array.
-    pub const fn to_bytes_le(&self) -> [u8; 32] {
-        todo!()
+    pub fn to_bytes_le(&self) -> [u8; 32] {
```
LucasLvy, line 78: can you add in the rust doc how fast is this function

crates/stark-felt/src/lib.rs
```diff
@@ -26,140 +37,193 @@ pub struct FromBytesError;
...
+        self.0.to_bytes_le()
     }
 
     /// Converts to big-endian bit representation.
-    pub const fn to_bits_be(&self) -> BitArray<BitArrayStore> {
-        todo!()
+    pub fn to_bits_be(&self) -> BitArray<BitArrayStore> {
```
LucasLvy, line 83: can you add in the rust doc how fast is this function

crates/stark-felt/src/lib.rs
```diff
@@ -26,140 +37,193 @@ pub struct FromBytesError;
...
+            .try_into()
+            .unwrap();
+
+        BitArray::new(limbs)
     }
 
     /// Converts to little-endian bit representation.
-    pub const fn to_bits_le(&self) -> BitArray<BitArrayStore> {
-        todo!()
+    pub fn to_bits_le(&self) -> BitArray<BitArrayStore> {
```
LucasLvy, line 100: can you add in the rust doc how fast is this function

LucasLvy (changes requested):
Can you make sure that it's no std compatible (i see some std imports).
More generally this crate aims to be used by everyone so can you please guide the devs to use the most performant functions plz (if they are all as performant as each other can you state it also).

crates/stark-felt/src/lib.rs
```diff
@@ -26,140 +37,193 @@ pub struct FromBytesError;
...
+    pub fn inverse(&self) -> Option<Self> {
+        Some(Self(self.0.inv()))
     }
 
     /// Finds the square root. There may be 2 roots for each square, and the lower one is returned.
-    pub const fn sqrt(&self) -> Option<Self> {
-        todo!()
+    pub fn sqrt(&self) -> Option<Self> {
+        let (root_1, root_2) = self.0.sqrt()?;
+        Some(Self(if root_1 < root_2 { root_1 } else { root_2 }))
```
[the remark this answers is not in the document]
fmoletta: Sounds good!

fmoletta:
> Can you make sure that it's no std compatible (i see some std imports).
> More generally this crate aims to be used by everyone so can you please guide the devs to use the most performant functions plz (if they are all as performant as each other can you state it also).

Hello! Thanks for the review!
Regarding no-std: I found some imports were missing for no-std so I added them in (41d8e81), and made sure I could compile a no-std project importing stark-felt. I couldn't find any explicit std-imports not gated behind the "std" feature, could you expand on this?
Regarding the comments on the `arbitrary` module: I added some small comments indicating that they are used to generate values for proptests. Is this clear enough or should I add examples / a link to the proptest documentation? This is also a private module, so end users shouldn't have to worry about it.
Regarding performance of certain methods: I ran some micro-benchmarks and wasn't able to find any substancial difference between them. I think it's a good idea to add benchmarks to the crate, so end users can gauge the performance themselves instead of relying on the documentation

crates/stark-felt/src/lib.rs
```diff
@@ -26,140 +37,193 @@ pub struct FromBytesError;
...
     }
 
     /// Converts to big-endian bit representation.
-    pub const fn to_bits_be(&self) -> BitArray<BitArrayStore> {
-        todo!()
+    pub fn to_bits_be(&self) -> BitArray<BitArrayStore> {
+        let mut limbs = self.0.representative().limbs;
+        limbs.reverse();
+
+        #[cfg(not(target_pointer_width = "64"))]
```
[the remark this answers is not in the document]
fmoletta: (264bd77)

Mirko-von-Leipzig:
> Regarding performance of certain methods: I ran some micro-benchmarks and wasn't able to find any substancial difference between them. I think it's a good idea to add benchmarks to the crate, so end users can gauge the performance themselves instead of relying on the documentation

I think the comment was more referring to which of those representations is actually the same as the inner type. i.e. does the field element store big endian or little endian? Or is it some other representation internally. While this does "leak" the inner workings a bit, it might be meaningful for those chasing to save instructions.

For example, in pathfinder we store felts as big endian in the database only because our field element uses big endian under the hood so its a no-op to convert.

Whether or not this actually matters is of course debatable :)

crates/stark-felt/src/lib.rs
```diff
@@ -382,22 +486,55 @@ mod formatting {
 
     /// Represents [Felt] in decimal by default.
     impl fmt::Display for Felt {
```
spapinistarkware, line 488: Should we display decimal or hexadecimal here?

crates/stark-felt/src/lib.rs
```diff
@@ -382,22 +486,55 @@ mod formatting {
...
+            let mut buf = [0u8; 4 * 20];
+            let mut i = buf.len() - 1;
+            let mut current = self.0.representative();
+            let ten = UnsignedInteger::from(10_u16);
+
+            loop {
+                let digit = if current < ten {
+                    current.limbs[3] as u8
+                } else {
+                    (current.div_rem(&ten).1).limbs[3] as u8
```
spapinistarkware, line 503: Why div_rem twice?

spapinistarkware (changes requested):

crates/stark-felt/src/lib.rs
```diff
@@ -26,140 +44,201 @@ pub struct FromBytesError;
...
+    pub fn pow(&self, exponent: u128) -> Self {
+        Self(self.0.pow(exponent))
     }
 
     /// Modular multiplication.
-    pub const fn mul_mod(&self, _rhs: &Self, _p: &Self) -> Self {
-        todo!()
+    pub fn mul_mod(&self, rhs: &Self, p: &Self) -> Self {
+        Self(FieldElement::from(
+            &(self.0 * rhs.0)
```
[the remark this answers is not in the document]
fmoletta: @xJonathanLEI has a use case for it in [starknet-rs](https://github.com/xJonathanLEI/starknet-rs/blob/ed14f483bb5d5171e6775960a24ee07018f3ff7c/starknet-crypto/src/ecdsa.rs#L88-L118).
I left it as it is part of the initial api, but we could discuss removing it and adding a `mod` method that specifically states that it performs modulo a number smaller than prime to support those use cases. This would also remove the need for `inverse_mod`.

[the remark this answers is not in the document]
spapinistarkware: I find this confusing and a potential trap. If noone opposed, I say remove.

crates/stark-felt/src/lib.rs
fmoletta: Fixed (2799d21)

crates/stark-felt/src/lib.rs
fmoletta: We have `LowerHex` & `UpperHex` implementations so it makes sense to display decimal here imo

crates/stark-felt/src/lib.rs
```diff
@@ -325,52 +404,77 @@ mod arithmetic {
...
 #[cfg(feature = "serde")]
 mod serde {
-    use ::serde::{Deserialize, Serialize};
+    use core::fmt;
+
+    use ::serde::{de, Deserialize, Serialize};
 
     use super::*;
 
     impl Serialize for Felt {
```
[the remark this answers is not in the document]
fmoletta: Fixed (61c553b)

crates/stark-felt/src/lib.rs
```diff
@@ -26,140 +44,201 @@ pub struct FromBytesError;
...
-        todo!()
+    /// This is as performant as [to_bits_le](Felt::to_bits_le)
+    pub fn to_bits_be(&self) -> BitArray<BitArrayStore> {
+        let mut limbs = self.0.representative().limbs;
+        limbs.reverse();
+
+        #[cfg(not(target_pointer_width = "64"))]
+        // Split limbs to adjust to BitArrayStore = [u32; 8]
+        let limbs: [u32; 8] = limbs
+            .map(|n| [(n >> 32) as u32, n as u32])
```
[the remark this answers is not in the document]
fmoletta: Fixed (29df244)

crates/stark-felt/src/lib.rs
```diff
@@ -325,52 +404,77 @@ mod arithmetic {
...
-            todo!()
+            Felt(self.0.neg())
         }
     }
 
     impl iter::Sum for Felt {
-        fn sum<I: Iterator<Item = Self>>(_iter: I) -> Self {
-            todo!()
+        fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
+            iter.fold(Self::ZERO, |augend, addend| augend + addend)
```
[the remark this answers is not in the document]
fmoletta: Fixed (148d086)

crates/stark-felt/src/lib.rs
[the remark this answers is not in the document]
fmoletta: Removed (5fc8a5a)

crates/stark-felt/src/lib.rs
```diff
@@ -26,140 +44,201 @@ pub struct FromBytesError;
...
+    pub fn to_bytes_le(&self) -> [u8; 32] {
+        self.0.to_bytes_le()
     }
 
     /// Converts to big-endian bit representation.
-    pub const fn to_bits_be(&self) -> BitArray<BitArrayStore> {
-        todo!()
+    /// This is as performant as [to_bits_le](Felt::to_bits_le)
+    pub fn to_bits_be(&self) -> BitArray<BitArrayStore> {
+        let mut limbs = self.0.representative().limbs;
```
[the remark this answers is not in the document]
fmoletta: @LucasLvy what do you think about this?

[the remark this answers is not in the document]
LucasLvy: If it changes it should be updated but IDK if it could change so much except if you do a big refacto on the felt type ? @fmoletta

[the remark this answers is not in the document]
xJonathanLEI: IMO even if it changes the worst that can happen is that downstream applications see a performance drop if they bump dependency and didn't notice (or just didn't lock dependencies). They still do compile. They would notice if they really care about performance anyways.

If we really want to, we can always make a semver-incompatible release so that Cargo doesn't auto upgrade.

spapinistarkware (approved):

PR #3, opened 2023-07-27
ci: update ci + add ensure no std

LucasLvy:
# Pull Request type

- Bugfix
- Feature
- Code style update (formatting, renaming)
- Build-related changes
- Documentation content changes
- Testing
- Other (please describe):

Resolves: #NA

## What is the new behavior?

-
-
-

## Other information

- - -
This change is [Reviewable](https://reviewable.io/reviews/starknet-io/types-rs/3)

### DIFF
--- a/crates/stark-felt/src/lib.rs
+++ b/crates/stark-felt/src/lib.rs
@@ -12,10 +12,10 @@ pub type BitArrayStore = [u64; 4];
 pub type BitArrayStore = [u32; 8];
 
 #[cfg(not(feature = "std"))]
-extern crate alloc;
+pub extern crate alloc;
 #[cfg(not(feature = "std"))]
 use alloc::string::ToString;
-#[cfg(not(feature = "std"))]
+#[cfg(all(not(feature = "std"), not(target_pointer_width = "64")))]
 use alloc::vec::Vec;
 
 use lambdaworks_math::{
@@ -418,9 +418,10 @@ mod arithmetic {
 
 #[cfg(feature = "serde")]
 mod serde {
-    use core::fmt;
-
     use ::serde::{de, ser::SerializeSeq, Deserialize, Serialize};
+    #[cfg(not(feature = "std"))]
+    use alloc::{format, string::String};
+    use core::fmt;
 
     use super::*;
 
@@ -475,6 +476,7 @@ mod serde {
 }
 
 mod formatting {
+
     use core::fmt;
 
     use super::*;
@@ -566,13 +568,15 @@ mod errors {
 
 #[cfg(test)]
 mod test {
-    use core::ops::Shl;
-
+    #[cfg(not(feature = "std"))]
+    use super::alloc::{format, string::String, vec::Vec};
     use crate::arbitrary::nonzero_felt;
+    use core::ops::Shl;
 
     use super::*;
 
     use proptest::prelude::*;
+    #[cfg(feature = "serde")]
     use serde_test::{assert_de_tokens, assert_ser_tokens, Configure, Token};
 
     proptest! {
@@ -879,6 +883,7 @@ mod test {
     }
 
     #[test]
+    #[cfg(feature = "serde")]
     fn deserialize() {
         assert_de_tokens(&Felt::ZERO, &[Token::String("0x0")]);
         assert_de_tokens(&Felt::TWO, &[Token::String("0x2")]);
@@ -892,6 +897,7 @@ mod test {
     }
 
     #[test]
+    #[cfg(feature = "serde")]
     fn serialize() {
         assert_ser_tokens(&Felt::ZERO.readable(), &[Token::String("0x0")]);
         assert_ser_tokens(&Felt::TWO.readable(), &[Token::String("0x2")]);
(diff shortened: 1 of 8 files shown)


spapinistarkware (dismissed):
> Reviewed 1 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starknet-io/types-rs/3)* status: 1 of 8 files reviewed, 1 unresolved discussion (waiting on @abdelhamidbakhta, @dan-starkware, @LucasLvy, and @xJonathanLEI)

___
*[`crates/stark-felt/src/lib.rs` line 422 at r1](https://reviewable.io/reviews/starknet-io/types-rs/3#-NauVHsn5FcpdZAWt67N:-NauVHsn5FcpdZAWt67O:b-ls74zq) ([raw file](https://github.com/starknet-io/types-rs/blob/b966a13c430980b7fed7dc43012a63e9664f9cfb/crates/stark-felt/src/lib.rs#L422)):*
> ```XML
> mod serde {
>     use ::serde::{de, ser::SerializeSeq, Deserialize, Serialize};
>     #[cfg(not(feature = "std"))]
> ```

Why is it gated? Why not always use the alloc one?

abdelhamidbakhta (approved):
lgtm

ensure_no_std/src/main.rs
```diff
@@ -0,0 +1,21 @@
...
+#[no_mangle]
+pub extern "C" fn _start() -> ! {
+    loop {}
+}
+
+#[global_allocator]
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+
+#[allow(unused_imports)]
+use stark_felt;
```
tdelabro, line 21: use `starknet-types-core` and `starknet-types-rpc ` too.
Otherwise, they are not actually part of the binary and checked

crates/stark-felt/src/lib.rs
```diff
@@ -11,11 +11,10 @@ pub type BitArrayStore = [u64; 4];
...
 pub type BitArrayStore = [u32; 8];
 
-#[cfg(not(feature = "std"))]
-extern crate alloc;
-#[cfg(not(feature = "std"))]
+pub extern crate alloc;
+
 use alloc::string::ToString;
-#[cfg(not(feature = "std"))]
+#[cfg(not(target_pointer_width = "64"))]
```
tdelabro, line 17: Why this?

crates/stark-felt/src/lib.rs
```diff
@@ -11,11 +11,10 @@ pub type BitArrayStore = [u64; 4];
 #[cfg(not(target_pointer_width = "64"))]
 pub type BitArrayStore = [u32; 8];
 
-#[cfg(not(feature = "std"))]
-extern crate alloc;
-#[cfg(not(feature = "std"))]
+pub extern crate alloc;
```
tdelabro, line 14: Alloc is always imported? In `std` too?

tdelabro (changes requested):

crates/stark-felt/src/lib.rs
LucasLvy: Check what Shahar said. Anyway the std lib just re exports everything from alloc I'm fine with both

crates/stark-felt/src/lib.rs
LucasLvy: Because it wasn't the right feature ?

ensure_no_std/src/main.rs
LucasLvy: For now they're empty

tdelabro:
@spapinistarkware 

> Why is it gated? Why not always use the alloc one?

We could because the std is just a reexport of the alloc.
This way allows for the `std` code not to change a bit, and just adds the `no_std` one ontop of it. 
But we could go the other way.

xJonathanLEI (changes requested):
Unless I'm missing something, I'm not convinced that this PR makes things `no_std` compatible. These two things are diffferent:

1. making sure the crate _ITSELF_ does not use anything from `std`;
2. making sure the crate _and its dependencies_ do not use anything from `std`.

This PR only does the former, whereas we actually need the latter.

Since the `no_std` attribute _only ensures the former_, the only way to test the latter (AFAIK) is to build the crate against some target where `std` support is just missing, which is what I originally added in the CI. This PR changes the CI to not do that, so CI is all green now, but the crate is actually _NOT_ `no_std`-compatible.

xJonathanLEI:
I could be wrong, but essentially this PR just invalidates the check that _actually_ ensures `no_std` compatibility.

LucasLvy:
The ensure no_std crate will compile the crate (and therefore all the dependencies) and if one of them uses something of std it panics.
What you added is a bare metal target which doesn't have an allocator so we can't use the alloc crate which is allowed in a wasm context

xJonathanLEI:
> The ensure no_std crate will compile the crate (and therefore all the dependencies) and if one of them uses something of std it panics.

This is not true. Again, `no_std` only checks the _current_ crate. Dependencies are free to use `std` as they wish and Rust wouldn't complain.

> What you added is a bare metal target which doesn't have an allocator

That's not true. `alloc` is available in that target. You can compile the entire `starknet-core` crate for that target. `starknet-core` uses `alloc` a lot.

xJonathanLEI:
OK I looked a bit deeper. So `ensure_no_std` relies on the panic handler not `no_std` to function.

That might work then, but I'm not sure. Need to check a bit more.

xJonathanLEI (approved):
Assuming that `ensure_no_std` works, I guess this looks good then.

tdelabro (approved):

PR #8, opened 2023-10-18
Implement starknet_core_types::curve module

pefontana:
# Implement starknet_core_types::curve module
* Move the Felt implementation to starknet_core_types
* Implement Implement starknet_core_types::curve module:
  * affine_point
  * projective_point 
  * curve_errors

Yes


nils-mathieu:
I'm not sure whether moving the `Felt` type to the core crate is a good idea. If we have `starknet-types-rpc` depend on it, it will bring a whole range of types that regular Starknet clients just don't care about.

`Felt` might not be the only type for which this matters though.

The question to be asked is whether the "rpc types" crate should depend on the "core types" crate. It's not clear to me.

Obvious pro is simplicity. Less crates to worry about = less time auditing dependencies and thinking about dependency trees. Obvious con is bringing a whole range of types that people my not need. Maybe just using feature flags a bit aggressively might suffice to solve the issue though.

I don't have a strong opinion on that.

crates/starknet-types-core/src/curve/affine_point.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    /// Returns the `x` coordinate of the point.
+    pub fn x(&self) -> Felt {
+        Felt(*self.0.x())
+    }
+
+    /// Returns the `y` coordinate of the point.
+    pub fn y(&self) -> Felt {
+        Felt(*self.0.y())
+    }
```
tdelabro, line 31: I expect the signature to be `pub fn x(&self) -> &Felt`.
This can be achieved with some raw pointers and a pinch of unsafe code.
Here is a code snippet summarizing how:
```rust
struct W1(u32);

impl W1 {
    fn inner(&self) -> &u32 {
        &self.0
    }
}

struct W2(u32);

struct W3(W1);

impl W3 {
    fn inner(&self) -> &W2 {
     let raw_pointer = self.0.inner() as *const u32 as *const W2;
     unsafe { &*raw_pointer }
    }
}
```

crates/starknet-types-core/src/curve/projective_point.rs
```diff
@@ -0,0 +1,212 @@
...
+        Self(ShortWeierstrassProjectivePoint::new([x.0, y.0, z.0]))
+    }
+
+    /// The point at infinity.
+    pub fn identity() -> ProjectivePoint {
+        Self(ShortWeierstrassProjectivePoint::neutral_element())
+    }
+
+    /// Creates the same point in affine coordinates. That is,
+    /// returns [x / z: y / z: 1] where `self` is [x: y: z].
```
tdelabro, line 25: `[x / z: y / z: 1]`

I don't understand this notation.
Do the brackets mean an array?
Does `/` mean division or separation? 
Does imply an operation, an equality or is it just a separation? 

Maybe it's standard and I'm just not well-versed enough in mathematics?

crates/starknet-types-core/src/curve/projective_point.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    /// Returns the `y` coordinate of the point.
+    pub fn y(&self) -> Felt {
+        Felt(*self.0.y())
+    }
+
+    /// Returns the `z` coordinate of the point.
+    pub fn z(&self) -> Felt {
+        Felt(*self.0.z())
+    }
```
tdelabro, line 46: Same as previous. We can return a reference with a bit of unsafe code

pefontana:
> I'm not sure whether moving the `Felt` type to the core crate is a good idea. If we have `starknet-types-rpc` depend on it, it will bring a whole range of types that regular Starknet clients just don't care about.
> 
> `Felt` might not be the only type for which this matters though.
> 
> The question to be asked is whether the "rpc types" crate should depend on the "core types" crate. It's not clear to me.
> 
> Obvious pro is simplicity. Less crates to worry about = less time auditing dependencies and thinking about dependency trees. Obvious con is bringing a whole range of types that people might not need. Just using feature flags a bit aggressively might suffice to solve the issue though.
> 
> I don't have a strong opinion on that.



great @nils-mathieu 
As we talked on telegram, I added a feature flag with the curve implementation
[15bed72](https://github.com/starknet-io/types-rs/pull/8/commits/15bed72041e4d908d0fb74e63d9b4e4d1ad16eeb)

I think it is okay to leave these implementations inside the starknet-types-core crate since I dont think there are plans to add many more structs inside the crates. In case there is any problem I can move all to a new starknet-arithmetic

crates/starknet-types-core/src/curve/affine_point.rs
pefontana: Hi guys!
I am not really a fan of adding unsafe code in such core libraries.
Returning the struct instead of the reference is not such a problem, so I think are best choice will be to leave it like that.
What do you think?

crates/starknet-types-core/src/curve/projective_point.rs
pefontana: It means division. That with the point (x, y, z) we create the new point
(x * inv_z, y * inv_z, 1)

where z_inv = multiplicative inverse of z

pefontana: I made some modifications in the documentation so it is clearer 
[971bef6](https://github.com/starknet-io/types-rs/pull/8/commits/971bef699bf35419fa7bb5ea55d12fe1b8958000)

crates/starknet-types-core/src/curve/affine_point.rs
nils-mathieu: I personally don't mind unsafe code as long as it's properly commented. However, for something like that, it's not clear whether it would be make a difference or not.

Here's what's to consider according to me:

When returning the type by value, Rust has to allocate 32 bytes on the stack, potentially clobbering some registers while trying to copy the return value into the allocated space. That's generally not good for performance.

When returning a reference, only eight bytes are needed, and they won't probably even be allocated at all because eight bytes fits in a regular CPU register. Moreover, because the function takes a reference in the first place, the function becomes essentially a no-op for the compiler. That's good for performance.

However, simply by marking the function as `#[inline(always)]` or `#[inline]`, the compiler will probably manage to do that optimization anyway,  removing the copy if possible. I know it struggles to do that with huge types (like 256 bytes), but we're well bellow that threshold.

To conclude: unsafe code would make the function fast always. With no unsafe, it's up to the compiler.

nils-mathieu:
Of note, if we do end up using unsafe code to return references, I'd say we should create a private function on those lines:

```rust
#[repr(transparent)]
struct Felt(FieldElement< /* ... */ >);

impl Felt {
    // no-op function to turn a field element reference into a `Felt`
    #[inline(always)]
    fn wrap(inner: &FieldElement< /* ... */ >) -> &Self {
        // SAFETY:
        //  `Felt` is a `#[repr(transparent)]` wrapper around `FieldElement< /* ... */ >`,
        //  which ensures that its memory layout is always the same as its content.
        unsafe { &*(inner as *const _ as *const Self) }
    }
}
```

crates/starknet-types-core/src/curve/affine_point.rs
tdelabro: >  I am not really a fan of adding unsafe code in such core libraries.

On the other hand, performances are utterly important in such core libraries. This method will be called millions of times in any program that uses the lib. We have to do our best the lib consumers.

Oppen (approved):

crates/starknet-types-core/Cargo.toml
```diff
@@ -1,13 +1,31 @@
...
+
+arbitrary = { version = "1.3.0", optional = true, default-features = false }
+num-traits = { version = "0.2.16", default-features = false }
+
+[features]
+default = ["std", "serde", "curve"]
+curve = []
+std = []
+alloc = ["serde?/alloc"]
+arbitrary = ["std", "dep:arbitrary"]
```
nils-mathieu, line 27: Honestly at this point if we do end up gating things behind feature flags, we should have a feature flag for `felt` as well.

Right now it wouldn't be very useful because everything depends on it, but we could imagine some types in the future that wouldn't.

Can you think of any type/feature that wouldn't depend on felt? I don't know the ecosystem well enough to think of one. If you can it would be wise to gate `Felt` as well (and enable that feature by default).

crates/starknet-types-core/src/curve/affine_point.rs
nils-mathieu: I think the consensus right now (after discusing about this on telegram), is that unsafe code should generally be avoided for this kind of case.

@tdelabro you convined me that it's a good idea to use it though. So after I have some data to back it up (a project with an example where this matters), I'll come back make a pull request.

For now it's perfectly fine to just go with copies, for the small improvement we will gain.

crates/starknet-types-core/src/curve/mod.rs
```diff
@@ -0,0 +1,3 @@
+pub mod affine_point;
+pub mod curve_errors;
+pub mod projective_point;
```
nils-mathieu, line 3: Not sure how we want to expose the module tree, but I think simply exporting those modules directly would be bettre API-wise.

```rust
mod affine_point;
mod curve_errors;
mod projective_point;

pub use self::affine_point::*;
pub use self::projective_point::*;
pub use self::curve_point::*;
```

Something like that.

crates/starknet-types-core/src/curve/affine_point.rs
```diff
@@ -0,0 +1,45 @@
+use crate::curve::curve_errors::CurveError;
+use crate::felt::Felt;
+use lambdaworks_math::cyclic_group::IsGroup;
+use lambdaworks_math::elliptic_curve::short_weierstrass::curves::stark_curve::StarkCurve;
+use lambdaworks_math::elliptic_curve::short_weierstrass::point::ShortWeierstrassProjectivePoint;
+use lambdaworks_math::elliptic_curve::traits::FromAffine;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AffinePoint(pub(crate) ShortWeierstrassProjectivePoint<StarkCurve>);
```
LucasLvy, line 9: Can you add a short description of this type plz

crates/starknet-types-core/src/curve/projective_point.rs
```diff
@@ -0,0 +1,213 @@
...
+use crate::felt::Felt;
+use core::ops;
+use lambdaworks_math::cyclic_group::IsGroup;
+use lambdaworks_math::elliptic_curve::short_weierstrass::curves::stark_curve::StarkCurve;
+use lambdaworks_math::elliptic_curve::short_weierstrass::point::ShortWeierstrassProjectivePoint;
+use lambdaworks_math::elliptic_curve::traits::EllipticCurveError::InvalidPoint;
+use lambdaworks_math::unsigned_integer::traits::IsUnsignedInteger;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ProjectivePoint(pub(crate) ShortWeierstrassProjectivePoint<StarkCurve>);
```
LucasLvy, line 12: can you add a short description of this type plz

LucasLvy (changes requested):
Still not convinced that we should move the felt type out of the felt crate, for now we don't have anything else. Can we re asses when we add something else ? For now can you plz put it back in the felt crate and add a `curve` feature flag

tdelabro:
Even if we don't do the unsafe stuff, I think we should still add `#[repr(transparent)]` to the wrappers types.

pefontana:
> Can you add a `curve` feature flag plz

Hi @LucasLvy !
The feature flag is already added
https://github.com/starknet-io/types-rs/pull/8/files#diff-5fc4f24b82d50a8341933bd608f0a88b530790f6a4a38d9c8d0583646bf5762fR2

crates/starknet-types-core/Cargo.toml
pefontana: For the plans we have in the starknet-core-types crate, almost everything depends on the Felt. I find it hard to imagine a user that would import this crate without the Felt type.
Do you have some examples?

pefontana:
> df43f6c

Sure @tdelabro 
Done [df43f6c](https://github.com/starknet-io/types-rs/pull/8/commits/df43f6cd65f58dfbc9ecbd75c58afe2c94efbb1e)

crates/starknet-types-core/src/curve/affine_point.rs
pefontana: Done!
[3a20aed](https://github.com/starknet-io/types-rs/pull/8/commits/3a20aed7cfed518017454a2b24c393c101cb3189)

nils-mathieu (approved):
Looks good to me.

tdelabro (approved):

PR #9, opened 2023-10-23
Implementing a code generator for `starknet-types-rpc`

nils-mathieu:
This pull request introduces a code generator (hosted in `/crates/starknet-types-rpc/generator`) that parses OpenRPC documents provided by Starkware ([link](https://github.com/starkware-libs/starknet-specs)) in order to automatically produce the required types.

I'm not sure whether you guys had something in mind for this, but either way I can help. I'm putting this draft PR here so you can check it out and determine whether this is of interest or not.

I was writing the generator anyway for my own personal project (specifically because none of the solutions currently available on <crates.io> perfectly fit my needs.

## Pull Request type

This is obviously a new feature :)

The crate is empty and exists as a placeholder.

## What is the new behavior?

The crate is no longer empty and contains type definitions!

Technically, yes. If someone depends on the crate and has something like:

```Rust
use starknet_types_rpc::*
use some_other_crate::BlockWithTxHashes;
```

They will get a compilation error 👀 

I hope no one is depending on this because I'm too lazy to write a migration guide.

## Steps

- [x] Allow param types to correctly deserialize from arrays and maps (that's part of the JSON-RPC specification).
- [x] Add optional support for `#![no_std]` (currently it's always required).
- [x] Remove the debug paths from the generated file.

## Unresolved Questions

Do we want to create a module per version of the specification? (i.e. v0_1_0, v0_2_0, etc.) It would probably be more correct but it would make the crate a bit harder to use (though it's probably not that bad).

## Other information

There obviously is some unresolved questions to be answered. The first one is whether code generation is the right approach for this project in the first place. I personally like writing everything by hand to have the freedom to add as much documentation as I want. I also generally like to handcraft everything. However, the following considerations made me change my mind:

- With a generator, adapating to API updates and new versions becomes trivial.

- Adding a feature flag to gate something specific provided by the crate is easy and fast. Easy to maintain as well.

- Creating submodules for every version of the API is very easy and requires little to no change to the generator.

- We avoid making mistakes while trying to follow the specification. Having a generator doesn't fix this directly, when one fix in the generator fixes a lot more than one fix on a single thing.

However, I do think there are some drawbacks to this:

- Using a generator makes it a bit harder for new contributor because they need to learn how it work instead of directly interacting with the types they know.

- Anything that's not directly in the specification (e.g. external types such as `stark-felt::Felt`) must be added as a special case in a separate configuration file.

- We're kinda bound to the spec. It's harder to create custom types and definitions. The data model of JSON is very peculiar and its direct translation to Rust can be a bit weird at times. Getting this right procedurally with a generator is definitely non trivial.

In any case, I'd be down to work on that. Generator or not.


nils-mathieu:
I just force-pushed to remove the commits related to the generator implementation.

The reason for that is that I found myself writing a lot of non-starknet related code and I decided to shift the generator to a separate crate that's general to any OpenRPC document. It's available [here](https://github.com/nils-mathieu/openrpc-gen).

nils-mathieu:
I'll put this ready for review for now, but I still need to write some tests to ensure that types are properly serialized/deserialized.

nils-mathieu:
# Unresolved Questions

Do we want to create a module per version of the specification? (i.e. `v0_1_0`, `v0_2_0`, etc.) It would probably be more correct but it would make the crate a bit harder to use (though it's probably not that bad).

LucasLvy:
@xJonathanLEI already has a code generator for rpc types https://github.com/xJonathanLEI/starknet-jsonrpc-codegen. Also i think that beerus already tried the openrpc type generation and it wasn't that good (can't remember why though i'll ask). For now my first impression is that it adds a lot of useless types and requires to hold the spec file in the repo which i don't like

nils-mathieu:
> Already has a code generator for rpc types

Yeah, I've talked tiny bit with them about it. He's the one that convinced me that it was better than writing everything by hand 😅.

The main thing was that it allowed him to find a few bugs in node implementations thanks to strict code generation, which I think is a nice thing to have.

For the other points, I discused in the PR post.

I like what they did but there several points:

1. It uses `serde_as` and a bunch of other dependencies that I didn't want to bring into the generator.
2. It misses a few quality-of-life detections that I think are important if anyone needs to maintain the repository (not the generator) later.

> it adds a lot of useless types and requires to hold the spec file in the repo which i don't like.

What useless types are you talking about? You kinda need to declare all the fields that will be serialized in one way or another, so there can't really be any useless fields.

There's some optimization you can do to reduce boilerplate, such as flattening structs that are used once, or those that only have one field (which are performed by my generator).

There's also a few special-case for `BlockId` and `SyncingStatus` because their "rusty" representation can be greatly improved if done by hand.

> requires to hold the spec file in the repo

It's true right now because the spec is kinda broken. For example it misses a few `"type": "object"` that would be needded to follow the JSON-Schema specification. I fixed those two typos in the spec and saved the fixed version in the repo. If that's something we don't want, we can always notify them of the issue.

Or we can just not save it and fix it every time. That's not a big deal it takes about 15 seconds.

crates/starknet-types-rpc/src/block_id.rs
```diff
@@ -0,0 +1,67 @@
+use serde::{Deserialize, Deserializer, Serialize};
```
tdelabro, line 1: Why are the types in this file manually created and not generated?

crates/starknet-types-rpc/src/generated/v0_5_0.rs
```diff
@@ -0,0 +1,1476 @@
...
+    // #/methods/starknet_getTransactionStatus/result/_anon/execution_status
+    #[serde(default)]
+    pub execution_status: Option<TxnExecutionStatus>,
+    // #/methods/starknet_getTransactionStatus/result/_anon/finality_status
+    pub finality_status: TxnStatus,
+}
+
+/// Parameters of the `starknet_specVersion` method.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SpecVersionParams {}
```
tdelabro, line 1321: Why is this one empty?

crates/starknet-types-rpc/src/generated/v0_5_0.rs
```diff
@@ -0,0 +1,1476 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BlockHashAndNumberParams {}
+
+/// Parameters of the `starknet_chainId` method.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ChainIdParams {}
+
+/// Parameters of the `starknet_syncing` method.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SyncingParams {}
```
tdelabro, line 1460: And those? Is it just that this RPC route does not take any params?

crates/starknet-types-rpc/src/generated/v0_5_0.rs
nils-mathieu: The `starknet_specVersion` RPC call takes no parameters.

crates/starknet-types-rpc/src/generated/v0_5_0.rs
nils-mathieu: Yep exactly

crates/starknet-types-rpc/src/block_id.rs
nils-mathieu: `BlockId` is used a lot and its actual representation is:

```rust
enum BlockId {
    Tag(BlockId),
    Hash(BlockIdHash),
    Number(BlockIdNumber),
}

struct BlockIdHash {
    block_hash: BlockHash,
}

struct BlockIdNumber {
    block_number: BlockNumber, // u64
}
```

which is not ideal because it's used a lot and it's a hassle to construct.

For this reason, I created a custom implementation of `Serialize` and `Deserialize` to make it just

```rust
enum BlockId {
    Tag(BlockTag),
    Hash(BlockHash),
    Number(BlockNumber),
}
```

For `SyncingStatus`, it's roughly the same.

It's "normal" representation is:

```rust
enum SyncingStatus {
    False(bool), // must always be `false`
    Syncing(SyncState),
}
```

Which I replaced by:

```rust
enum SyncingState {
    NotSyncing, // serializes as `false`, deserializes only `false
    Syncing(SyncState),
}
```

This is mainly for convenience and to reduce boilerplate in downstream crates.

nils-mathieu: Not that I know of, but feel free to tell if you see any of them!

xJonathanLEI:
My only comment here would be that the generator should live in a separate repo but I see that's already the case now.

Generally, I like the approach of generating model code, which is why I did it for `starknet-rs`. I do also agree that `serde_as` is suboptimal and I plan to get rid of it at some point (nothing planned at the moment though).

However, I skimmed through the code and there doesn't seem to be any support for query-only version? It's not actually part of the specs so it makes sense that the directly generated code doesn't support it yet. That said, it's widely used in practice and need to be supported for the code to be useful.

LucasLvy:
What's the status here ? Waiting for review ?

nils-mathieu:
> What's the status here ? Waiting for review ?

Yes!

LucasLvy:
ok after you fix all the no std imports i'll review

crates/starknet-types-rpc/src/custom_serde.rs
```diff
@@ -0,0 +1,137 @@
...
+        }
+        *cursor.next().unwrap() = b'x';
+        *cursor.next().unwrap() = b'0';
+
+        let remaining = cursor.len();
+
+        // SAFETY:
+        //  We only wrote ASCII characters to the buffer, ensuring that it is only composed
+        //  of valid UTF-8 code points.
+        let s = core::str::from_utf8(&buffer[remaining..]).unwrap();
```
LucasLvy, line 40: why bother doing all that instead of just 
```rs
format!("{self:x"}");
```

crates/starknet-types-rpc/src/custom_serde.rs
```diff
@@ -0,0 +1,137 @@
...
+                let Some(digits) = v.strip_prefix("0x") else {
+                    return Err(E::custom("expected a hexadecimal string starting with 0x"));
+                };
+
+                let mut n = 0u64;
+                for b in digits.bytes() {
+                    let unit = match b {
+                        b'0'..=b'9' => b as u64 - b'0' as u64,
+                        b'a'..=b'f' => b as u64 - b'a' as u64 + 10,
+                        b'A'..=b'F' => b as u64 - b'A' as u64 + 10,
```
LucasLvy, line 71: Any reason not to do that ?
```suggestion
                        b'a'..=b'f' | b'A'..=b'F' => b as u64 - b'a' as u64 + 10,
```

LucasLvy (changes requested):
Mini review, can you add comments in your code plz, this repo will be the base for the whole ecosystem, i expect that a lot of people will read this code so can you add some detailed comments on what you're doing inside each function it'd be very helpful

crates/starknet-types-rpc/src/custom_serde.rs
nils-mathieu: To avoid to things:

1. Having to gothrough the Rust `fmt::Arguments` type which uses dynamic dispatch (meaning it cannot be optimized by `rustc`)
2. Havint to allocate a whole `String` every time a number has to be serialized (which is very common).

crates/starknet-types-rpc/src/custom_serde.rs
LucasLvy: ah indeed read too fast

nils-mathieu:
> Mini review, can you add comments in your code plz, this repo will be the base for the whole ecosystem, i expect that a lot of people will read this code so can you add some detailed comments on what you're doing inside each function it'd be very helpful

Done! I've added a whole bunch of comments to describe how the algorithm works, as well as some further optimizations on the deserialization side.

LucasLvy (approved):

PR #10, opened 2023-10-24
Improve existing documentation for the `stark-felt` crate.

nils-mathieu:
This pull request provide some additional documentation to the methods defined for the `Felt` type.

# Important Notes

One of the examples I made for `inverse_mod` do not pass. Here is the bogus example:

```rust
use stark_felt::{Felt, NonZeroFelt};
let a = Felt::from(3);
let p = Felt::from(5).try_into().unwrap();

let b = a.inverse_mod(&p).unwrap();
assert_eq!(a.mul_mod(&b, &p), Felt::ONE);
```

This do not pass.

Apparently, `b` is computed to be `4` instead of `2`.

I'll open an issue for that.

# State

Blocked by #11 .

### DIFF
--- a/crates/stark-felt/src/lib.rs
+++ b/crates/stark-felt/src/lib.rs
@@ -29,7 +29,7 @@ use lambdaworks_math::{
         element::FieldElement, fields::fft_friendly::stark_252_prime_field::Stark252PrimeField,
     },
     traits::ByteConversion,
-    unsigned_integer::{element::UnsignedInteger, traits::IsUnsignedInteger},
+    unsigned_integer::element::UnsignedInteger,
 };
 
 #[cfg(feature = "arbitrary")]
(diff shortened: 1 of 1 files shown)


crates/stark-felt/src/lib.rs
```diff
@@ -130,73 +161,223 @@ impl Felt {
...
+    ///
+    /// ```rust
+    /// # use stark_felt::Felt;
+    /// let a = Felt::from(3);
+    ///
+    /// assert_eq!(a.pow(2u32), Felt::from(9));
+    /// assert_eq!(a.pow(3u32), Felt::from(27));
+    /// assert_eq!(a.pow(0u32), Felt::ONE);
+    /// ```
+    pub fn pow(&self, exponent: impl IsUnsignedInteger) -> Self {
```
Oppen, line 289: This one would be a breaking change, and would also leak the lambdaworks dependency onto users.

nils-mathieu: Oooh makes sense. I'll revert it 'cause it obviously falls out of the scope of the PR.

I feared converting to `u128` might trash registers if the function is not properly optimized by rustc.

LucasLvy (approved):

tdelabro (approved):

tdelabro:
a bit outdated sorry

Issue #11, opened 2023-10-24
bug: `inverse_mod` returns invalid value


nils-mathieu:
# Bug Report

**stark-felt version:** 0.0.3
**commit**: [ea52b64](https://github.com/starknet-io/types-rs/commit/ea52b64a52fd832222adf400feb2dd3b4dc286a2)

## Behavior:

The `inverse_mod` functions returns an invalid value: `3.inverse_mod(5)` returns 4 instead of 2.

Indeed `3 * 4 == 12` and `12 % 5 == 2 != 1`.

## Other information:

I found this writing documentation (see #10). If you download that PR and run `cargo test`, you'll get the error.

Currently, the only test for `inverse_mod` is this one:

```rust
#[test]
fn inverse_mod_in_range(x in nonzero_felt(), p in nonzero_felt()) {
    let nzp = NonZeroFelt(p.0);
    prop_assert!(x.inverse_mod(&nzp) <= Some(Felt::MAX));
    prop_assert!(x.inverse_mod(&nzp) < Some(p));
}
```

Which only checks whether the resulting inverse is less than `p`. It should probably check something like `x.inverse_mod(&nzp).mul_mod(x, &nzp) == 1`.

nils-mathieu:
Update: this test do not pass.

```rust
#[test]
fn inverse_mod_in_range(x in nonzero_felt(), p in nonzero_felt()) {
    let nzp = NonZeroFelt(p.0);

    let Some(ret) = x.inverse_mod(&nzp) else { return Ok(()) };

    prop_assert!(ret <= Felt::MAX);
    prop_assert!(ret < p);
    prop_assert!(ret.mul_mod(&x, &nzp) == Felt::ONE);
}
```

pefontana:
Thanks for the report @nils-mathieu 
I am currently checking it!

pefontana:
Hi @nils-mathieu !
I just reviewed it, and the documentation of the methods are not clear.
The `inverse_mod` method first calculates the multiplicative inverse and then it calculates the mod. We don't calculate the inverse of the number changing the Prime field. The same happens with the `mul_mod` method, we first calculate the multiplication and then apply the mod. So, the condition `ret.mul_mod(&x, &nzp) == Felt::ONE` don't have to be true

I just checked and we don't use this operation in the cairo-vm, I am not really sure if other projects need them.
I can make a PR adding more documentation to the methods, or just delete them if they generate confusion

nils-mathieu:
[...]

PR #15, opened 2023-11-14
Reimplement Felt::mul_mod and Felt::mod_inverse

pefontana:
# Reimplement Felt::mul_mod and Felt::mod_inverse

Reimplements the Felt methods:
* Felt::mul_mod
* Felt::mod_inverse

- Feature

No

fixes : #11

### DIFF
--- a/crates/starknet-types-core/src/felt.rs
+++ b/crates/starknet-types-core/src/felt.rs
@@ -173,7 +173,7 @@ impl Felt {
         Self(self.0.pow(exponent.0.representative()))
     }
 
-    // TODO Add docs
+    // TODO Add docs,  Check name
     /// Modular multiplication between `self` and `rhs` modulo `p`.
     pub fn mul_mod(&self, rhs: &Self, p: &NonZeroFelt) -> Self {
         let multiplicand = BigInt::from_bytes_be(num_bigint::Sign::Plus, &self.to_bytes_be());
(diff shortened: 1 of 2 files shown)


crates/starknet-types-core/src/felt.rs
```diff
@@ -170,14 +173,47 @@ impl Felt {
...
+
+        let result = multiplicand.mul(multiplier).mod_floor(&modulus);
+
+        let (_, buffer) = result.to_bytes_be();
+        let mut result = [0u8; 32];
+
+        result[(32 - buffer.len())..].copy_from_slice(&buffer[..]);
+
+        // safe .unwrap()
+        Felt::from_bytes_be(&result).unwrap()
```
LucasLvy, line 192: Why not use felts and use `div_rem` to avoid (expensive?) conversions with bytes

LucasLvy (changes requested):

crates/starknet-types-core/src/felt.rs
pefontana: Because the idea of the methods is to operate with Field Elements of mod `p` instead of the Felt that operate in the p for the Starknet PRIME.
Right now we cant make this operations with lambdaworks, so we need to use BigInts

pefontana: I agree that the performance could be better but for a first implementation I think is good, also these two methods are not the most used.
In the future I will talk with the lambdaworks team to see how we can improve the performance

Oppen: Relevant info: the cited implementation is already relying on `num_bigint`, so at the very least this should not regress performance (we'll have to test eventually of course).

Oppen (approved):

juanbono (approved):

LucasLvy (approved):

PR #16, opened 2023-11-17
feat: felt creation from bytes is infallible

Oppen:
Expose 4 ways to create a field element from arbitrary bytes:
- `from_bytes_be`/`from_bytes_le`: take exactly 32 bytes and interpret them as a big/little endian number of 256 bits and then build a field element from it, modulo STARK prime (in whatever representation `Stark252PrimeField` uses, currently Montgomery);
- `from_bytes_be_slice`/`from_bytes_le_slice`: take any number of bytes and interpret them as a big/little endian number modulo STARK prime by decomposing into 256 bits integers and multiplying by the appropriate factors.

# Pull Request type

- Feature

Current behavior mimics that of `lambdaworks`, where depending on the length of the slice passed to `from_bytes_xe` (where `x` is either `b` or `l`) one of the following happens:
1. Length is exactly 32 bytes: conversion succeeds and the bytes are interpreted as a single 256 bits integer that is then stored modulo `P`;
2. Length is less than 32 bytes: conversion fails and an error is returned;
3. Length is more than 32 bytes: excess bytes are ignored, and 32 bytes are interpreted as a single 256 bits integer that is then stored modulo `P`.

## What is the new behavior?

Now the functionality is divided in two:
- One family for exactly 32 bytes;
- One family for arbitrary length, where:
  1. If length is exactly 32 bytes, we see the old behavior;
  2. If length is less than 32 bytes, we 0-pad the most significant bytes to fit 32 bytes;
  3. If length is more than 32 bytes, we split the number in 32 bytes chunks (0-padding when needed) and compute the number as `∑Di*2^i` where `Di` is the i-th "digit" of 512 bits units. This should be exactly equivalent to the following code (for little endian):
      ```rust
      let x = BigUint::from_bytes_le(bytes);
      let x = x % P;
      ```
      This way, we get the same result but without allocations on the heap. The powers of 2 are computed on demand by multiplying a constant.

Yes. The signature of the `from_bytes_xe` functions change, they no longer return an error.


crates/starknet-types-core/src/felt.rs
```diff
@@ -63,20 +63,104 @@ impl Felt {
...
         FieldElement::from_bytes_be(bytes)
             .map(Self)
-            .map_err(|_| FromBytesError)
+            .expect("from_bytes_be shouldn't fail for these many bytes")
     }
 
-    /// Creates a new [Felt] from its little-endian representation in a [u8] slice.
-    /// This is as performant as [from_bytes_be](Felt::from_bytes_be)
-    pub fn from_bytes_le(bytes: &[u8]) -> Result<Self, FromBytesError> {
+    pub fn from_bytes_le(bytes: &[u8; 32]) -> Self {
```
pefontana, line 72: here too

pefontana (approved):

fmoletta (approved):

LucasLvy (approved):

PR #18, opened 2023-11-28
Impl num traits

### LINKED ISSUE #7: dev: impl more num-traits traits
https://docs.rs/num-traits/latest/num_traits/ops/inv/trait.Inv.html
 https://docs.rs/num-traits/latest/num_traits/pow/trait.Pow.html
 https://docs.rs/num-traits/latest/num_traits/identities/trait.One.html

tdelabro:
resolves #7 

## What is the new behavior?

`num_traits::{Inv, Pow, One, Zero}` are implemented on `Felt`
They exist behind a feature flag. So if you don't need them, you can build the crate without.

No

### DIFF
--- a/crates/starknet-types-core/src/felt/mod.rs
+++ b/crates/starknet-types-core/src/felt/mod.rs
@@ -980,8 +980,8 @@ mod errors {
 #[cfg(test)]
 mod test {
     use super::alloc::{format, string::String, vec::Vec};
+    use super::felt_arbitrary::nonzero_felt;
     use super::*;
-    use crate::felt_arbitrary::nonzero_felt;
     use core::ops::Shl;
     use proptest::prelude::*;
     #[cfg(feature = "serde")]
--- a/crates/starknet-types-core/src/felt/mod.rs
+++ b/crates/starknet-types-core/src/felt/mod.rs
@@ -27,11 +27,8 @@ pub type BitArrayStore = [u64; 4];
 #[cfg(not(target_pointer_width = "64"))]
 pub type BitArrayStore = [u32; 8];
 
+#[cfg(any(test, feature = "alloc"))]
 pub extern crate alloc;
-
-use alloc::string::ToString;
-#[cfg(not(target_pointer_width = "64"))]
-use alloc::vec::Vec;
 
 use lambdaworks_math::{
     field::{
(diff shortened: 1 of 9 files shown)


crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -1,9 +1,17 @@
+#[cfg(test)]
+mod felt_arbitrary;
+
 use core::ops::{Add, Mul, Neg};
 
 use bitvec::array::BitArray;
 use num_bigint::BigInt;
+
+use num_traits::{One, Zero};
```
LucasLvy, line 9: shouldn't this be behind a feature flag ?

tdelabro: Here it's used not to be implemented on our Felt type, but to be able to call it's impl on `Bigint`.

Oppen (approved):

LucasLvy:
rebase

tdelabro:
@Oppen I also added a commit to fix the way the alloc feature was used. It was a bit weird, potentially a bug

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -219,15 +232,24 @@ impl Felt {
...
     pub fn to_bits_le(&self) -> BitArray<BitArrayStore> {
         let limbs = self.0.representative().limbs;
 
-        #[cfg(not(target_pointer_width = "64"))]
         // Split limbs to adjust to BitArrayStore = [u32; 8]
         let limbs: [u32; 8] = limbs
             .into_iter()
             .flat_map(|n| [n as u32, (n >> 32) as u32])
-            .collect::<Vec<u32>>()
+            .collect::<alloc::vec::Vec<u32>>()
```
LucasLvy, line 252: import it at the top plz
```suggestion
            .collect::<Vec<u32>>()
```

LucasLvy (changes requested):

crates/starknet-types-core/src/felt/mod.rs
tdelabro: Look line 244. This is only available if the feature `alloc` is enabled.
And it's not used anywhere except in those functions.
It's better to have it here, where it is used, already behind the feature flag

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -21,12 +27,9 @@ pub type BitArrayStore = [u64; 4];
 #[cfg(not(target_pointer_width = "64"))]
 pub type BitArrayStore = [u32; 8];
 
+#[cfg(any(test, feature = "alloc"))]
```
LucasLvy, line 30: why test ?

crates/starknet-types-core/src/felt/mod.rs
LucasLvy: ah right

LucasLvy (changes requested):
otherwise lgtm

crates/starknet-types-core/src/felt/mod.rs
tdelabro: coz alloc::vec::Vec is used to setup tests. But not in the "actual" code

LucasLvy (approved):

PR #21, opened 2023-12-06
refactor: Optimize Hexadecimal Serialization and Deserialization for u64

codeesura:
# Pull Request type

Before these changes, the `serialize` and `deserialize` methods for `u64` in the `NumAsHex` trait were implemented with less emphasis on performance optimization. The serialize method utilized a dynamic approach for string construction, and deserialize handled hexadecimal string parsing without specific optimization techniques. While fully functional, these implementations had potential areas for performance improvement, especially in scenarios involving large-scale or frequent serialization/deserialization operations.

Resolves: #NA

## What is the new behavior?

- Optimized the `serialize` method in the `NumAsHex` trait for `u64` to use a fixed-size buffer, enhancing performance and reducing memory overhead.
- Refined the `deserialize` method for improved efficiency in parsing hexadecimal strings into `u64`, using a more direct approach and safe overflow handling with `checked_mul` and `checked_add`.

No

## Other information

This PR focuses on enhancing performance without altering the functionality of the existing code. It has been thoroughly tested to ensure that the new implementations of `serialize` and `deserialize` maintain the same behavior as their predecessors.

### DIFF
--- a/crates/starknet-types-rpc/src/custom_serde/num_as_hex.rs
+++ b/crates/starknet-types-rpc/src/custom_serde/num_as_hex.rs
@@ -107,10 +107,14 @@ impl<'de> NumAsHex<'de> for u64 {
                 // Because we already checked the size of the string earlier, we know that
                 // the following code will never overflow.
                 let hex_bytes = &bytes[2..];
+
+                // Trim leading zeros from the hexadecimal part for efficient processing
                 let trimmed_hex = hex_bytes
                     .iter()
                     .skip_while(|&&b| b == b'0')
                     .collect::<Vec<_>>();
+
+                // Check if the significant part of the hexadecimal string is too long for a 64-bit number
                 if trimmed_hex.len() > 16 {
                     return Err(E::custom("hexadecimal string too long for a 64-bit number"));
                 }
--- a/crates/starknet-types-rpc/src/custom_serde/num_as_hex.rs
+++ b/crates/starknet-types-rpc/src/custom_serde/num_as_hex.rs
@@ -97,6 +97,15 @@ impl<'de> NumAsHex<'de> for u64 {
                     return Err(E::custom("expected a hexadecimal string starting with 0x"));
                 }
 
+                // Aggregate the digits into `n`,
+                // Digits from `0` to `9` represent numbers from `0` to `9`.
+                // Letters from `a` to `f` represent numbers from `10` to `15`.
+                //
+                // As specified in the spec, both uppercase and lowercase characters are
+                // allowed.
+                //
+                // Because we already checked the size of the string earlier, we know that
+                // the following code will never overflow.
                 let hex_bytes = &bytes[2..];
                 let mut n = 0u64;
                 for &b in hex_bytes {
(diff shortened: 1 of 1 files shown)


LucasLvy:
Multiple things, don't remove the useful documentation, sure it makes the code less complex but without a benchmark it's hard to say optimize

crates/starknet-types-rpc/src/custom_serde/num_as_hex.rs
```diff
@@ -128,16 +106,20 @@ impl<'de> NumAsHex<'de> for u64 {
...
+                        b'A'..=b'F' => 10 + b - b'A',
                         _ => return Err(E::custom("invalid hexadecimal digit")),
                     };
-
-                    n = n * 16 + unit;
+                    n = n
+                        .checked_mul(16)
+                        .ok_or_else(|| E::custom("integer overflowed 64-bit"))?
+                        .checked_add(digit as u64)
+                        .ok_or_else(|| E::custom("integer overflowed 64-bit"))?;
```
Oppen, line 122: Since invalid characters are already checked at this point, and given we're trying to optimize, just a length check before the loop should suffice for overflow checks: 64-bit numbers take at most 16 hexadecimal digits + 2 for the `0x`, so just make sure `hex_bytes.len() <= 16` and use the unchecked variants here.

codeesura: Thank you for your suggestion. I've updated the code to use `skip_while` to skip leading zeros in hexadecimal strings, counting only the significant digits.

nils-mathieu:
I am not sure how this modification makes the serialization routines faster. They are doing the same exact thing, except without using iterators.

In the deserialization function, you add a `.collect<Vec<_>>()` which copies part of the input to the heap. That's bad because allocations are generally slow, and in the context of parsing a number (which is very common), that overhead cannot be ignored. That allocation can be removed, sure, but then it's the same as for the serialization routine: the code does exactly the same thing, just without iterators.

If you *really* need to optimize those functions, I think the best is to junp down to unsafe code and access the string without checking bounds within the loops, and avoid checking whether the final string is valid UTF-8 (we know it is because we wrote ascii characters to it).

That being said, is that gain really worth introducing unsafe code into the codebase?

Oppen:
> If you _really_ need to optimize those functions, I think the best is to junp down to unsafe code and access the string without checking bounds within the loops, and avoid checking whether the final string is valid UTF-8 (we know it is because we wrote ascii characters to it).

The bounds checks are likely optimized out if you check the length beforehand. What may be more interesting here is replacing `String::from_utf8` by `String::from_utf8_unchecked` and, as you mentioned, skipping the copy altogether when skipping zeros by working with slices. [str::trim_start_matches](https://doc.rust-lang.org/std/primitive.str.html#method.trim_start_matches) is probably what we want, although a more manual approach may or may not be faster.

In any case, without a benchmark this is all a bit moot. I'm in favor of it if it simplifies the code, not so much if we claim it's faster, make it more complex, and provide no evidence of speed improvement.

Oppen:
Also, if we have no benchmark I would much rather just use `from_str_radix` and let the stdlib work for us.

nils-mathieu:
> Also, if we have no benchmark I would much rather just use `from_str_radix` and let the stdlib work for us.

Ahah, now that I think about it, using stdlib might even be faster than what I had wrote 😅

PR #26, opened 2023-12-19
feat: hex formatting helper

Oppen:
Little helper to format felts as hex strings with prefix notation. Because the method returns a `String`, the `alloc` features is required.

# Pull Request type

- Feature

Yes

### DIFF
--- a/crates/starknet-types-core/src/felt.rs
+++ b/crates/starknet-types-core/src/felt.rs
@@ -200,6 +200,12 @@ impl Felt {
         BitArray::new(limbs)
     }
 
+    /// Helper to produce a hexadecimal formatted string.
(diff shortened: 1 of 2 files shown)


crates/starknet-types-core/src/felt.rs
```diff
@@ -200,6 +200,13 @@ impl Felt {
         BitArray::new(limbs)
     }
 
+    /// Helper to produce a hexadecimal formatted string.
+    /// Equivalent to calling `format!("{self:#x}")`.
+    #[cfg(feature = "alloc")]
```
LucasLvy, line 205: this is also ok for std feature no?

LucasLvy (approved):

PR #35, opened 2024-01-12
Add support for the v0.6 of the Starknet API

nils-mathieu:
This pull request adds support for the 0.6 version of the specification.

Also it removes config files from the exported package, to avoid forcing everybody to download them when they get the crate from crates.io


nils-mathieu:
wait was there a rebase? Why are those commits not found

Oooh, no. I did forget to pull.

LucasLvy (approved):

LucasLvy (changes requested):
Upgrade the package version to the rpc version plz

LucasLvy (approved):

PR #42, opened 2024-02-26
feat: impl ToBigInt and ToBigUInt

tdelabro:


LucasLvy (changes requested):
tests

pefontana (changes requested):
Hi @tdelabro !
I dont get why we need this, when we have the Felt methods ` pub fn to_bigint(&self) -> BigInt` and `pub fn to_biguint(&self) -> BigUint`

tdelabro:
@pefontana we need it in order to be compatible with traits that would require the implementation of ToBigInt and ToBigUInt. 
The whole num-traits thing is behind a feature flag, but if you need it to interface with some libs methods or traits, you can enable the flag and use the felt type there.

crates/starknet-types-core/src/felt/num_traits_impl.rs
```diff
@@ -1,6 +1,19 @@
 use super::Felt;
+use num_bigint::{ToBigInt, ToBigUint};
 use num_traits::{FromPrimitive, Inv, One, Pow, ToPrimitive, Zero};
 
+impl ToBigInt for Felt {
+    fn to_bigint(&self) -> Option<num_bigint::BigInt> {
```
Oppen, line 6: Are there infallible variants for this trait? It looks like it would be more ergonomic to use given all `Felt`s can be converted.

tdelabro: No there is not. This is the way this lib trait is defined so that virtually any type can implement it.

LucasLvy: Add a comment in the doc to say that it’s safe to unwrap ?

tdelabro: done

tdelabro:
> still waiting for tests

The conversion logic is already tested there: https://github.com/starknet-io/types-rs/blob/36f5be1e849b60d60697b873e428836305429ca6/crates/starknet-types-core/src/felt/mod.rs#L1931

There is no logic to be tested in the code added by this PR.

LucasLvy:
for the current implementation yes but if we change something we'll get rekt, just copy paste the other tests idk

tdelabro:
@Oppen can we merge this?

maciejka (approved):

Oppen (approved):
LGTM

Oppen:
> @Oppen can we merge this?

LGTM but I'm not blocking the merge. Also I don't think I have merge rights anymore, official maintainers are now only Levy and Peter AFAIK.

That said, I do see a point in the tests Lucas asked for, mostly to detect regressions. Even if the tests only assert that the two ways to convert give the same results that's some assurance for future us. I do disagree that the difference it makes here is enough to block merging tho.

LucasLvy (approved):

PR #44, opened 2024-03-08
Add double implementation for Felt

tcoratger:
# Pull Request type

- Feature

Add double implementation for Felt

## What is the new behavior?

- This pull request introduces a utility method `double` for the `Felt` type. The `double` method allows for more convenient and concise code when needing to double a `Felt` value. Previously, one would have to write `x + x` to achieve the same result, which can be verbose and less readable. With the addition of `double`, users can now simply call `x.double()` to achieve the same outcome.

No

### DIFF
--- a/crates/starknet-types-core/proptest-regressions/felt/mod.txt
+++ b/crates/starknet-types-core/proptest-regressions/felt/mod.txt
@@ -1,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc 7cc7e05316cf3a4fc6fc302a9d0addd5ec76e6cd733ef30c109c14be6d0bf9d5 # shrinks to x = Felt(FieldElement { value: UnsignedInteger { limbs: [300729720636178304, 1831347307108920996, 11337269826263459472, 8939133563709308857] } })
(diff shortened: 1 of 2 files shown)


crates/starknet-types-core/proptest-regressions/felt/mod.txt
```diff
@@ -0,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
```
LucasLvy, line 1: remove that

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -360,6 +360,11 @@ impl Felt {
         Self(self.0.square())
     }
 
+    /// Doubles the point `self`
+    pub fn double(&self) -> Self {
+        Self(self.0.add(self.0))
```
LucasLvy, line 365: why do we need that ?

LucasLvy (changes requested):

crates/starknet-types-core/proptest-regressions/felt/mod.txt
tcoratger: Done

crates/starknet-types-core/src/felt/mod.rs
LucasLvy: can you benchmark `felt + felt` vs `Felt::TWO * felt` then  plz?

tcoratger: If you want I can no problem at all :) but I'm 90% sure because in finite field arithmetic, multiplication is much heavier than the addition due to the complexity of the operation, we did some similar tests here: https://github.com/xJonathanLEI/starknet-rs/pull/547 with significant improvements when using `a + a` instead of `Felt::TWO * a`

LucasLvy: Ok let's do it
https://github.com/lambdaclass/lambdaworks/tree/main/math/src/field#-benchmarks

LucasLvy (approved):

PR #47, opened 2024-03-25
perf(hash): create pedersen hasher in a const

LucasLvy:
Pedersen hasher creation is slow and not needed to be done at every call of the `hash` function so put that in a const to improve the performances.

### DIFF
--- a/crates/starknet-types-core/Cargo.toml
+++ b/crates/starknet-types-core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "starknet-types-core"
-version = "0.0.9"
+version = "0.0.10"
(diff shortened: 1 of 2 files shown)


juanbono (approved):

tdelabro:
Couldn't this be done at `const` time? Given some changes in the upstream lib

Oppen (approved):

LucasLvy:
It could but we would need to wait for a release which we don't want to. We'll change to that as soon as it's available

pefontana (approved):

pefontana:
Thanks @LucasLvy !

PR #48, opened 2024-03-26
Add Pedersen benchmark

pefontana:
# Add Perdersen benchmark
* Add a criterion benchmark that runs the Pedesen::Hash function
* Add a little Makefile to the starknet-core-types crate
* Update Rust toolchain to 1.74.0 in `starknet-types-core` and `ensure_no_std` crates

## Does this introduce a breaking change?
No

PR #49, opened 2024-03-26
perf: update lambdaworks to 0.6.0

Oppen:
Updates lambdaworks to 0.6.0 to be able to use the optimized Pedersen hash.

This includes a compile-time optimization to reduce the time used to initialize the hash, and arithmetic optimizations to reduce the time of the hash itself.

NO ❌ 

## Other information

Merge after the benchmarks PR #48


LucasLvy (approved):
Can you format plz otherwise lgtm

PR #53, opened 2024-03-28
Add `new_unchecked` method for `AffinePoint`

tcoratger:
# Pull Request type

- Feature

## What is the new behavior?

- Add `new_unchecked` method for `AffinePoint`

No

### DIFF
--- a/crates/starknet-types-core/src/curve/affine_point.rs
+++ b/crates/starknet-types-core/src/curve/affine_point.rs
@@ -18,7 +18,7 @@ impl AffinePoint {
         )?))
     }
 
-    pub fn new_unchecked(x: Felt, y: Felt) -> AffinePoint {
+    pub const fn new_unchecked(x: Felt, y: Felt) -> AffinePoint {
         Self(ShortWeierstrassProjectivePoint::new([
             x.0,
             y.0,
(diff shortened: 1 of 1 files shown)


crates/starknet-types-core/src/curve/affine_point.rs
```diff
@@ -18,6 +18,14 @@ impl AffinePoint {
         )?))
     }
 
+    pub fn new_unchecked(x: Felt, y: Felt) -> AffinePoint {
```
LucasLvy, line 21: Can we make this function const ?

tcoratger: Yes sure, done

crates/starknet-types-core/src/curve/affine_point.rs
```diff
@@ -18,6 +18,14 @@ impl AffinePoint {
         )?))
     }
 
+    pub const fn new_unchecked(x: Felt, y: Felt) -> AffinePoint {
```
LucasLvy, line 21: What’s the behavior if you send incorrect values ? Does it panic ? Can you add a small description over the function explaining that plz

tcoratger: Just updated the documentation here. 

As per what is done here in the same context for Arkworks https://github.com/arkworks-rs/algebra/blob/bb663bc27e776af81e66c94d9016bde2dbba9968/ec/src/models/short_weierstrass/group.rs#L120, this method does not check if the point is on the curve or not.

As its name suggests, the method must be used with caution and full knowledge of the facts for very specific use cases.

LucasLvy (approved):

PR #57, opened 2024-04-09
feat: poseidon and pedersen extended benchmarks

maciejka:
Add benchmarks for poseidon and pedersen hashes for both `hash` and `hash_array` functions.
There are three benchmarks for each hash function:
* `*_hash_single_input` - same random pair of inputs is used for each `hash` function timing 
* `*_hash_multiple_inputs` - different random pair of inputs is used for each `hash` function timing, probably gives more realistic measurements
* `*_hash_array` - for input lengths: 2, 4, 8, 16, different  values are used for each `hash_many` timing


crates/starknet-types-core/benches/criterion_hashes.rs
```diff
@@ -0,0 +1,108 @@
...
+fn poseidon_hash_single_input(c: &mut Criterion) {
+    let mut rng = ChaCha8Rng::seed_from_u64(SEED);
+    let mut group = c.benchmark_group("Poseidon");
+
+    let (x, y) = rnd_pair_of_felts(&mut rng);
+
+    group.bench_function("hash - single input", |bench| {
+        bench.iter(|| black_box(Poseidon::hash(&x, &y)))
+    });
+}
```
LucasLvy, line 34: not sure how useful is this one

crates/starknet-types-core/benches/criterion_hashes.rs
```diff
@@ -0,0 +1,108 @@
...
+        );
+    });
+}
+
+fn poseidon_hash_array(c: &mut Criterion) {
+    let mut rng = ChaCha8Rng::seed_from_u64(SEED);
+
+    let mut group = c.benchmark_group("Poseidon");
+
+    for n in [2, 4, 8, 16].iter() {
```
LucasLvy, line 54: I think we need to test with larger input sizes
```suggestion
    for n in [10, 100, 1000, 10000].iter() {
```

LucasLvy, line 54: I think we need to test with larger input sizes
```suggestion
    for n in [10, 100, 1000, 10000].iter() {
```

crates/starknet-types-core/benches/criterion_hashes.rs
```diff
@@ -0,0 +1,108 @@
...
+use starknet_types_core::felt::Felt;
+use starknet_types_core::hash::{Pedersen, Poseidon};
+use starknet_types_core::hash::StarkHash;
+
+const SEED: u64 = 3;
+
+fn rnd_felt(rng: &mut ChaCha8Rng) -> Felt {
+    let mut felt: [u8; 32] = Default::default();
+    rng.fill_bytes(&mut felt);
+    Felt::from_bytes_be(&felt)
```
LucasLvy, line 14: this will most likely only generate big numbers, we prob also want to add a random number `0 < x <= 32` to generate "more" random numbers

LucasLvy (changes requested):
comments hold both for poseidon and pedersen. Also can you change the `Makefile` so we can run the benchmark using it?

crates/starknet-types-core/benches/criterion_hashes.rs
maciejka: Eventually not very useful, I wanted to keep both single and multiple versions in order to eliminate measuring benchmarking artifacts.

maciejka: Btw, there is a 25% difference between single and multiple versions:
image
I attribute it to the fact the single version probably benefits from some low level caching. Happy to hear an alternative explanation.

crates/starknet-types-core/benches/criterion_hashes.rs
maciejka: Out of curiosity, what is the use case for hashing 10000 long vec of felts?

crates/starknet-types-core/benches/criterion_hashes.rs
maciejka: Right!

crates/starknet-types-core/benches/criterion_hashes.rs
juanbono: I think it makes sense since calculating the Sierra class hash for a given Sierra file involves doing `Poseidon(SierraProgramFelts)`. Depending on the contract it can be a huge vector of felts.

crates/starknet-types-core/benches/criterion_hashes.rs
LucasLvy: I think i'd remove it then

crates/starknet-types-core/benches/criterion_hashes.rs
LucasLvy: can be up to 81k for class hash
Also the transaction hash requires a hash chain for the calldata which can be up to 4k
https://docs.starknet.io/documentation/tools/limits_and_triggers/
so def add the 10k plz

maciejka: done

crates/starknet-types-core/benches/criterion_hashes.rs
maciejka: done

LucasLvy (approved):

PR #74, opened 2024-06-18
feat: impl TryInto<primitive> for Felt

tdelabro:


crates/starknet-types-core/src/felt/primitive_conversions.rs
```diff
@@ -0,0 +1,330 @@
...
+
+            // Let's call `s` the `size_of_type`.
+            //
+            // There is 3 ways the value we are looking for can be encoded in the Felt bytes:
+            // 1. Positive numbers:
+            //  Bytes [s..] are all set to zero, and byte [s] value must not exceed 127
+            // 2. Negative numbers:
+            //  Bytes [s..] should match `MINUS_TWO_BYTES_REPR`, and byte [s] value must not be less than 128
+            // 3. -1:
+            //  Bytes are those of Felt::MAX
```
LucasLvy, line 122: i don't understand that. why 128 and 127 when this macro is used for all the types

LucasLvy (changes requested):

crates/starknet-types-core/src/felt/primitive_conversions.rs
```diff
@@ -0,0 +1,330 @@
...
+            false => Felt::ZERO,
+        }
+    }
+}
+
+impl From<Felt> for bool {
+    fn from(value: Felt) -> Self {
+        value != Felt::ZERO
+    }
+}
```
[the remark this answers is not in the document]
tdelabro: Indeed in Rust there is no impl of `From<u8> for bool` and `as` is not allowed either.
We should manually do `value != Felt::ZERO` instead

crates/starknet-types-core/src/felt/primitive_conversions.rs
tdelabro: i8::MAX 127i8    == Felt::ZERO + 127
i8::MIN -128i8 == Felt::ZERO - Felt::from(128) ==  Felt::MAX - Felt::from(127) == Felt::MAX - Felt::from(1) - Felt::from(126)

 Felt::MAX - Felt::from(1) is Felt::from(-2) and it's bytes repr is:
 ```rust
 const MINUS_TWO_BYTES_REPR: [u8; 32] = [
    255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
    255, 255, 255, 255, 255, 16, 0, 0, 0, 0, 0, 0, 8,
];
```

From there all the values will just change the last bytes, reducing it's value.
So if -2i8 is encoded as 255, then
-3i8 will be 255 - 1 == 254
-12i8 will be 255 - 10 == 245
and -128 will be 255 - 126 == 129

Our computers use the `one's complement` system to represent negative numbers as bytes.
Which means that `-2i8.to_le_bytes() == [254]`, which is `255 -1`, which is the value of the last byte in `MINUS_TWO_BYTES_REPR -1`.
So to get our `-2i8` value from the bytes of `MINUS_TWO_BYTES_REPR` we just have to do `i8::from_le_bytes(MINUS_TWO_BYTES_REPR[s-1]) - 1`.
Let's see:
For `-2`: `i8::from_le_bytes([255u8; 1]) -1 == -1i8 -1 == -2`
For `i8::MIN`: `i8::from_le_bytes([129; 1]) -1 == -127i8 -1 == -128`

Now for the ` && bytes_le[size_of_type - 1] >= 0x80`.
for `i8` it's logic, if I take a value like `Felt::from(-200)`, it's byte rep will be `[57, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 16, 0, 0, 0, 0, 0, 0, 8]` and if which qualify for our ` // Negative numbers` branch. So we will get `offsetted_value == 57` and therefore parse Felt::form(-200) as `57i8`. So we need this additional check.

Now let's do the same for `i16`:
- `MIN` repr is `[1, 128, 255, ..]`,  this is parsed as `-32767`, we subtract 1 and we get `-32768`, the value we were looking for.
- `MIN -1` repr is `[0, 128, 255, ..]`, this is parsed as `-32768` and returns an error when we try the checked subtraction, which is what we want.
- `MIN -2` repr is `[255, 127, 255, ..]`, this is parsed as `32767`, we subtract 1 and get `32766` and we are fucked.

We can easily protect ourselves from this by adding the ` && bytes_le[size_of_type - 1] >= 0x80` check

File: crates/starknet-types-core/src/felt/primitive_conversions.rs (hunk 2 above)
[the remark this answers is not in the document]
Comment (tdelabro) [reply crates/starknet-types-core/src/felt/primitive_conversions.rs]: done

File: crates/starknet-types-core/src/felt/primitive_conversions.rs [hunk 3]
```diff
@@ -0,0 +1,334 @@
...
+                    && bytes_le[size_of_type - 1] <= 0x7f
+                {
+                    Ok(<$into>::from_le_bytes(
+                        bytes_le[..size_of_type].try_into().unwrap(),
+                    ))
+                }
+                // Negative numbers
+                else if bytes_le[size_of_type..] == MINUS_TWO_BYTES_REPR[size_of_type..]
+                    && bytes_le[size_of_type - 1] >= 0x80
+                {
```
LucasLvy, line 132:
```suggestion
            fn try_from(value: Felt) -> Result<Self, Self::Error> {
                // We'll split the conversion in 3 case:
                // Felt >= 0
                // Felt == -1
                // Felt < -1

                // Get the size of the type we'll convert the felt into
                let size_of_type = core::mem::size_of::<$into>();
                // Convert the felt as bytes
                let bytes_le = value.to_bytes_le();

                // Case 1: Felt >= 0
                // Our felt type can hold values up to roughly 2**252 bits which is encoded in 32 bytes.
                // The type we'll convert the value into can hold `size_of_type` (= { 1, 2, 4, 8, 16 }) bytes.
                // If all the bytes after the last byte that our target type can fit are 0 it means that the
                // number is positive.
                // The target type is a signed type which means that the leftmost bit of the last byte
                // (remember we're in little endian) is used for the sign (0 for positive numbers)
                // so the last byte can hold a value up to 0b01111111
                if bytes_le[size_of_type..].iter().all(|&v| v == 0)
                    && bytes_le[size_of_type - 1] <= 0b01111111
                {
                    // We take the `size_of_type` first bytes because they contain the useful value.
                    Ok(<$into>::from_le_bytes(
                        bytes_le[..size_of_type].try_into().unwrap(),
                    ))
                }
                // Case 2: Felt == -1
                // This is the little endian representation of Felt::MAX. And Felt::MAX is exactly -1
                // [
                //    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
                //    0, 0, 0, 0, 0, 17, 0, 0, 0, 0, 0, 0, 8,
                // ]
                // The only valid Felt with those most significant bytes is Felt::MAX.
                // All other arrays with those most significant bytes are invalid Felts because they would be >= Prime.
                else if bytes_le[24..] == [17, 0, 0, 0, 0, 0, 0, 8] {
                    return Ok(-1);
                }
                // Case 3: Felt < -1
                // Similarly to how we checked that the number was positive by checking the bytes after the
                // `size_of_type` byte we check that all the bytes after correspond to the bytes of a negative felt
                // The leftmost bit is use for the sign, as it's a negative number it has to be 1.
                else if bytes_le[size_of_type..] == MINUS_TWO_BYTES_REPR[size_of_type..]
                    && bytes_le[size_of_type - 1] >= 0b10000000
                {
                    // We take the `size_of_type` first bytes because they contain the useful value.

```

tdelabro: done

LucasLvy (approved):
lgtm

tdelabro:
@pefontana or @juanbono we need a second reviewer

tdelabro:
@juanbono @pefontana @maciejka @Oppen we need a second reviewer!

Oppen (approved):

tdelabro:
@0xLucqs @maciejka, it looks like oppen review doesn't qualify for merging. Can you give me another that does?

pefontana (approved):

PR #76, opened 2024-06-21
Add Non-Human-Readable Deserialization for Felt Using Serde

jbcaron:
# Add Non-Human-Readable Deserialization for Felt Using Serde

## Pull Request type

- Feature

Currently, the `Felt` type does not support custom serialization and deserialization for `bincode` and other formats that differentiate between human-readable and binary forms.

## What is the new behavior?

- Implemented `Deserialize` for `Felt` to correctly parse hex strings and sequences of 32 bytes, ensuring compatibility with `bincode`.

No

### DIFF
--- a/crates/starknet-types-core/src/felt/mod.rs
+++ b/crates/starknet-types-core/src/felt/mod.rs
@@ -935,7 +935,11 @@ mod serde_impl {
         where
             D: ::serde::Deserializer<'de>,
         {
-            deserializer.deserialize_str(FeltVisitor)
+            if deserializer.is_human_readable() {
+                deserializer.deserialize_str(FeltVisitor)
+            } else {
+                deserializer.deserialize_seq(FeltVisitor)
+            }
         }
     }
 
@@ -959,6 +963,19 @@ mod serde_impl {
                 .map(Felt)
                 .ok_or(String::from("Expected hex string to be prefixed by '0x'"))
                 .map_err(de::Error::custom)
+        }
+
+        fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+        where
+            A: de::SeqAccess<'de>,
+        {
+            let mut bytes = [0u8; 32];
+            for (i, byte) in bytes.iter_mut().enumerate() {
+                *byte = seq
+                    .next_element()?
+                    .ok_or_else(|| de::Error::invalid_length(i, &"expected 32 bytes"))?;
+            }
+            Ok(Felt::from_bytes_be_slice(&bytes))
         }
     }
 }


LucasLvy (changes requested):
Can you add a few tests ?

LucasLvy (changes requested):
The length of the sequence is always 32? is it enforced somewhere ? if not can you add a test where the length is < 32 and one where it is > 32 ?

jbcaron:
I've implemented error handling for sequence lengths that are not exactly 32 bytes during deserialization. To ensure we provide accurate and efficient error messages, I considered two different approaches. I would appreciate your feedback on which approach is better.

### Approach 1:
```rust
if let Some(extra) = seq.size_hint() {
    if extra > 0 {
        return Err(de::Error::invalid_length(32 + extra, &"32 bytes"));
    }
}
```

- **Pros**: Uses `size_hint` to quickly estimate the number of extra elements, potentially reducing the number of iterations.
- **Cons**: `size_hint` might not always be accurate, and its behavior can vary based on the iterator implementation.

### Approach 2:
```rust
if seq.next_element::<u8>()?.is_some() {
    let mut count = 33;
    while seq.next_element::<u8>()?.is_some() {
        count += 1;
    }
    return Err(de::Error::invalid_length(count, &"32 bytes"));
}
```

- **Pros**: Directly iterates through elements to count extra ones, ensuring accurate length calculation.
- **Cons**: Might be less efficient due to additional iterations, especially with many extra elements.

Which approach do you think is more suitable for this case?

LucasLvy:
why do we want to enforce 32 bytes ? I think we should allow `<= 32`. I tend to prefer the 2nd approach as the 1st one is very unreliable

jbcaron:
> why do we want to enforce 32 bytes ? I think we should allow `<= 32`. I tend to prefer the 2nd approach as the 1st one is very unreliable

if we allow `<= 32`, I think it's better to use little-endian representation instead of big-endian

LucasLvy:
What would little endian add here ?

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -960,6 +964,26 @@ mod serde_impl {
...
+
+        fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+        where
+            A: de::SeqAccess<'de>,
+        {
+            let mut bytes = [0u8; 32];
+            for (i, byte) in bytes.iter_mut().enumerate() {
+                *byte = seq
+                    .next_element()?
+                    .ok_or_else(|| de::Error::invalid_length(i, &"32 bytes"))?;
```
tdelabro, line 976: For extra compactness, can't we accept sequences with fewer elements than 32?
Eg: Felt::ZERO could be serialized as `[0]`, rather than `[0; 32]`.
And all values under 255 on a single element, and so on.

tdelabro: We are kinda doing this already with the human-readable form. Felt::ZERO is `0x0` not `0x000000000..0000`

tdelabro: Oh I can see @LucasLvy already suggested this.
I agree with him

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -1661,236 +1685,46 @@ mod test {
...
-                Token::U8(0),
-                Token::U8(0),
-                Token::U8(0),
-                Token::U8(0),
-                Token::U8(0),
-                Token::U8(0),
-                Token::U8(0),
-                Token::U8(0),
+                Token::Seq { len: Some(2) },
+                Token::U8(1 | 0x80),
```
tdelabro, line 1689: Do we really need this first element containing len and the compressed flag?
can't we just encode `Felt::ONE` as
```
&[
Token::Seq { len: Some(1) },
Token::U8(1),
]
```
And know that because we are using big endian repr, missing bytes are `0`s and just be happy with it?

tdelabro (changes requested):

crates/starknet-types-core/src/felt/mod.rs
tdelabro: Serialization can be:

```rust
let be_bytes = self.to_bytes_be();
let mut i = match be_bytes.iter().find(|b| b != 0) {
 Some(i) => i,
 None => { // handle Felt::ZERO and return }
}
let mut seq = serializer.serialize_seq(Some(32 - i))?;
while i < 32 {
    seq.serialize_element(be_bytes[i])?;
    i += 1;
}
seq.end();
```

And deserialization:
```rust
let mut buffer = [0; 32];
let mut i = 32;
while i > 0 {
   if let Some(b) = seq.next_element()? {
         buffer[i-1] = b;
   } else {
           break;
   }
  i -= 1;
}
if i == 0 && seq.next_element()?.is_some() {
  // Return error invalid len
}
```

tdelabro: Why wouldn't this way simpler impl work?

jbcaron: Using `seq.next_element()?` to infer lengths and handle missing bytes as zeros isn’t reliable for all serializers. Some serializers may not support this implicit handling, leading to incorrect deserialization. Explicitly encoding the length and compressed flag ensures that the deserializer knows exactly how many bytes to expect and process, making the format self-descriptive and robust across different implementations.

tdelabro: But isn't what we are already doing with `Some(32 - i)`?

And how exactly could it fail?

tdelabro: https://docs.rs/serde/latest/serde/trait.Serializer.html#tymethod.serialize_seq

```
The argument is the number of elements in the sequence, which may or may not be computable before the sequence is iterated. Some serializers only support sequences whose length is known up front.
```

We are fulfilling all conditions described there

And nothing is said about what you are describing there: https://docs.rs/serde/latest/serde/de/trait.SeqAccess.html

Can you give examples or documentation to support your point?

jbcaron: It's a misunderstanding on my part.
By following your logic we can directly use se/deserialization using slice

### Serialization
```rust
let bytes = self.to_bytes_be();
let first_significant_byte = bytes.iter().position(|&b| b != 0).unwrap_or(32);
serializer.serialize_bytes(&bytes[first_significant_byte..])
```

### Deserialization
```rust
fn visit_bytes<E>(self, bytes: &[u8]) -> Result<Self::Value, E>
where
    E: de::Error,
{
    if bytes.len() > 32 {
        return Err(E::invalid_length(
            bytes.len(),
            &"a byte array of length up to 32",
        ));
    }
    Ok(Felt::from_bytes_be_slice(&bytes))
}
```

cchudant: The idea here we had with @jbcaron was to make a binary format that never exceeds 32 bytes in the worse case (abitrary hash), and compresses it down to a few bytes in the best cases (small felt)

To do that, we think of the input as a byte stream and we use the felt unused bits in the first byte to encode a flag telling us if the felt is compressed or not

When the felt is compressed we treat that first byte as the length, so we read `len` more bytes from the wire -
when it's not compressed, we treat it as the first byte of the hash and complete it by reading 31 mores bytes.

However, we realized there is an issue: using serde `serialize_seq` (or `serialize_bytes`), bincode will prepend the length of the sequence before the felt, and we end up with >32 bytes felts :(
This is not bincode's problem though it's fundamental to how serde works

Taking inspiration from fixed-sized arrays implementation in serde, they are serialized using `serialize_tuple`, which requires the length to be known beforehand... and we don't know the length before the first byte.

That gave me an idea though - which I tried here https://gist.github.com/cchudant/9bbb025c60a9fe23ca56b895fdcff13c
It works by treating the Felt as a tuple of (u8, \<rest\>) with \<rest\> being another tuple that is `DeserializeSeed`ed with the length
Nesting tuples like this has no effect on the actual bytes on the wire at the end, so we are free to do that
The resulting encoding never exceeds 32 bytes as intended and compressed stuff correctly 

@tdelabro what do you think of that?

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -936,11 +938,23 @@ mod serde_impl {
...
                 serializer.serialize_str(&format!("{:#x}", self))
             } else {
-                let mut seq = serializer.serialize_seq(Some(32))?;
-                for b in self.to_bytes_be() {
-                    seq.serialize_element(&b)?;
+                let bytes = self.to_bytes_be();
+                let first_significant_byte = bytes.iter().position(|&b| b != 0).unwrap_or(31);
+                if first_significant_byte > 1 {
+                    let len = 32 - first_significant_byte;
+                    let mut seq = serializer.serialize_seq(Some(len + 1))?;
```
tdelabro, line 945: Because we are already storing the size. We should call `serialize_eq` with `None` in order to spare the extra byte which contains the len at the beginning.
Otherway there is no use to store the len a second time ourself

tdelabro: It makes our worst case 33 bytes and our best 3 bytes (value, our len, their len). We want 32 and 2.

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -936,11 +938,23 @@ mod serde_impl {
             if serializer.is_human_readable() {
                 serializer.serialize_str(&format!("{:#x}", self))
             } else {
-                let mut seq = serializer.serialize_seq(Some(32))?;
-                for b in self.to_bytes_be() {
-                    seq.serialize_element(&b)?;
+                let bytes = self.to_bytes_be();
```
tdelabro, line 941: You can avoid the if else nesting of `if first_significant_byte > 1` doing the following:
```rust
let be_bytes = self.to_bytes_be();
let mut i = match be_bytes.iter().find(|b| b != 0) {
 Some(i) => i,
 None => { // handle Felt::ZERO and return }
}
let mut seq = serializer.serialize_seq(None)?;
if i != 0 {
  seq.serialize_element(32 - i | COMPRESSED)?;
}
while i < 32 {
    seq.serialize_element(be_bytes[i])?;
    i += 1;
}
seq.end();
```

crates/starknet-types-core/src/felt/mod.rs
tdelabro: Still, why do you need the flag for?
Get rid of it and if you have 32 elements, it's not compressed. If you have less it is compressed.
The flag doesn't add anything there

tdelabro (changes requested):

crates/starknet-types-core/src/felt/mod.rs
tdelabro: You need the byte that contain flag+len only when you serialize [Felt].
Then, because the size in bytes of a Felt can change you have to do:
[flag+len, byte, .., byte, flag+len, byte, flag+len, byte, .., byte, ...]

tdelabro: So my recomandation would be:
1. impl Serde on Felt the way I described
2. Create public method `serialize_sequence_of_felt_packed()` and `deserialize_sequence_of_felt_packed()` that can be called inside `#[serde(de/serialize_with = "path")]` and implement your self descriptive packed representation here

tdelabro:
@jbcaron do you plan on working this one forward?

Issue #81, opened 2024-07-16
bug: Panic on too big Felt


FabijanC:
# Bug Report

**types-rs version:**

0.1.5

**Current behavior:**

When using `Felt::from_hex` or deserializing, if the input is too long, the code panics.

**Expected behavior:**

I'd prefer it to return `Err` rather than panic because I can do nothing about that except write my own wrapper, which beats the purpose of having a unified Felt type.

**Steps to reproduce:**

E.g.:

```rust
Felt::from_hex(&format!("0x{}", "f".repeat(62)))
```

**Other information:**

FabijanC:
I updated the steps to reproduce with more info.

FabijanC:
Not opening a new issue because the following finding is relevant to this topic:

The `expecting` method of `FeltVisitor` is not implemented properly:
https://github.com/starknet-io/types-rs/blob/a6bcf4a5471d3673d23fda30fc55f7b333f4ec14/crates/starknet-types-core/src/felt/mod.rs#L963

It results in errors like this:
```
Invalid block ID: invalid type: number, expected Failed to deserialize hexadecimal string at line 1 column 33'
```

tdelabro:
This no longer panic for 65.
But, for 63 the value wraps. Which result in a felt with has a different value from the one that was given originally.

tdelabro:
@0xLucqs @gabrielbosio what do you think?
Mathematically speaking, it is not necessarily an error, as numbers bigger than the field's max value do indeed wrap.
BUT, this method already prevents you from using a hex string longer that the field max num of chars once hex serialized.

Which means that you can only successfully parse numbers biggers than Felt::MAX that are lower or equal to 0xfff...repeat 64 time...ff. Which is really quirky and weird.

I will add a check to prevent this

0xLucqs:
maybe we should have 2 functions, one that errors and another one that wraps ?

tdelabro:
closed by #166

PR #82, opened 2024-07-17
Fix FeltVisitor expecting message

FabijanC:
# Pull Request type

- Bugfix

Wrong message used in the `expecting` method of FeltVisitor, not following the documentation:
```
The message should complete the sentence "This Visitor expects to receive ...", for example the message could be "an integer between 0 and 64". The message should not be capitalized and should not end with a period.
```

Issue #81 makes several related reports, some of which have their root in the code imported from lambdaworks. If those parts of the issue are not going to be addressed in this repository, then the issue can be labeled as resolvable by this PR.

## What is the new behavior?

The fixed method now uses the expected message structure.

No

### DIFF
--- a/crates/starknet-types-core/src/felt/mod.rs
+++ b/crates/starknet-types-core/src/felt/mod.rs
@@ -960,7 +960,7 @@ mod serde_impl {
         type Value = Felt;
 
         fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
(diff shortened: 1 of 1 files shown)


Oppen (approved):

LucasLvy (approved):

tdelabro (approved):

tdelabro:
has been fixed in another PR changes

PR #86, opened 2024-07-31
Publish to creates.io workflow

maciejka:
Publish to creates.io workflow

# Pull Request type

- Build-related changes

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,25 +1,26 @@
-name: Publish to Cargo
-
 on:
   push:
-    branches: [ master ]
+    tags:        
+      - '*'           
+  workflow_dispatch:
+
+name: Publish
 
 jobs:
   publish:
+    name: Publish
     runs-on: ubuntu-latest
-
-    name: 'publish'
-
-    # Reference your environment variables
-    environment: cargo
-
     steps:
-        - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+  
+      - name: Install stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
```
0xLucqs, line 25: use `rustup show` like in the other workflows

.github/workflows/publish.yml
```diff
@@ -0,0 +1,29 @@
+on:
+  push:
+    branches:
+      - master
```
0xLucqs, line 4:
```suggestion
      - main
```

0xLucqs (changes requested):

.github/workflows/publish.yml
0xLucqs: unresolving as it's still not fixed ?

0xLucqs (changes requested):

0xLucqs (approved):

tdelabro (approved):

PR #89, opened 2024-08-21
Bugfix: binary deserialization `Felt`

xrvdg:
`Felt` supports both human readable and binary serialization, but the binary deserializer was missing. 

Yes, but likely no-one is affected by it as deserialization was not working. The (de)serialization is done with bytes rather than sequences. 

## Other information

The serialization and deserialization test have been combined.  Serialization and deserialization should not be tested separately as that makes it too easy to miss a case.

### DIFF
--- a/crates/starknet-types-core/src/felt/mod.rs
+++ b/crates/starknet-types-core/src/felt/mod.rs
@@ -959,9 +959,12 @@ mod serde_impl {
     impl<'de> de::Visitor<'de> for FeltVisitor {
         type Value = Felt;
 
-        // TODO(xrvdg) change this
         fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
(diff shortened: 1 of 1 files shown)


crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -1627,187 +1641,35 @@ mod test {
...
-            )],
-        );
-    }
+    fn serde() {
+        use serde_test::assert_tokens;
 
-    #[test]
-    #[cfg(feature = "serde")]
-    fn serialize() {
-        assert_ser_tokens(&Felt::ZERO.readable(), &[Token::String("0x0")]);
```
dan-starkware, line 1645: Please remove unused imports

dan-starkware (changes requested):

dan-starkware (approved):

0xLucqs (approved):

Issue #99, opened 2024-10-08
bug(felt): Felt is not zeroized on drop


Trantorian1:
# Bug Report

**types-rs version:** [7ef3023](https://github.com/starknet-io/types-rs/commit/7ef30234da7957f7585ea832d57d13083c76995c)

**Current behavior:** [`Felt`](https://github.com/starknet-io/types-rs/blob/7ef30234da7957f7585ea832d57d13083c76995c/crates/starknet-types-core/src/felt/mod.rs#L45) does not implement any zeroizing when it is dropped, opening the door for potential memory read attacks on sensitive information such as private keys.

**Expected behavior:**

`Felt` should implement [`#[derive(ZeroizeOnDrop)]`](https://docs.rs/zeroize/latest/zeroize/#custom-derive-support) or similar to allow memory zeroing when dropped. This could be feature-gated.

Given `Felt` is currently based on `lambdaworks-math` [`FieldElement`](https://github.com/lambdaclass/lambdaworks/blob/021c3b66de310aa7cc5af8e392cb958caadfae74/math/src/field/element.rs#L40), and this does not implement any zeroing logic, either a manual implementation using [`ptr::write_volatile`](https://doc.rust-lang.org/std/ptr/fn.write_volatile.html) and [`atomic::compiler_fence`](https://doc.rust-lang.org/std/sync/atomic/fn.compiler_fence.html) or an update to `lambdaworks-math` would be needed.

A relevant [issue](https://github.com/lambdaclass/lambdaworks/issues/927) has been opened there as well.

**Other information:**

While there certainly are more common ways to leak sensitive cryptographic information like private keys, it seems like for something in our control such as in-memory representation we should at least provide this as an option.

PR #100, opened 2024-10-09
feat(zeroize): added zeorizing to `Felt`

Trantorian1:
This is feature-gated behind the `zeroing` flag

# Pull Request type

- Feature

[`Felt`](https://github.com/starknet-io/types-rs/blob/7ef30234da7957f7585ea832d57d13083c76995c/crates/starknet-types-core/src/felt/mod.rs#L45) provides no option to zeroize its memory, which can be an issue when dealing with sensitive cryptographic information such as private keys.

Resolves: #99 

## What is the new behavior?

- Implemented zeroizing for `Felt` through the [`zeroize`](https://crates.io/crates/zeroize) crate. While `zeroize` is not used directly, Felt implements manually the `Zeroize` trait in the hopes of making this behavior more 'standard'. This has been gated behind the `zeroing` feature. 

> [!NOTE]
> Contrarily to `zeroize`, this is done using safe rust by leveraging `core::mem::take` to replace the in-memory representation of a felt with its default value of `Felt::ZERO`.

No. While the `zeroing` feature has been added to `default`, zeroing has not been implemented on drop, both because this would interfere with the `Copy` trait of `Felt` but also because it seems like it should be up to the end user to determine when they are dealing with sensitive information.


crates/starknet-types-core/Cargo.toml
```diff
@@ -44,6 +46,7 @@ serde = ["alloc", "dep:serde"]
 prime-bigint = ["dep:lazy_static"]
 num-traits = []
 papyrus-serialization = ["std"]
+zeroing = ["zeroize"]
```
tdelabro, line 49: zeroing = ["dep:zeroize"]
in ordr to disable the implicit feature.
https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

Currently you have two features available 
`zeroing` the one you defined
and 
`zeroize` from the default behaviour of `optional = true`

tdelabro, line 49: for this types of feature that just depend on one single crate, I tend to name the feature by the crate name. Here `zeroize` rather than `zeroing`. Wdyt?

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -1062,6 +1062,18 @@ mod errors {
     }
 }
 
+mod zeroing {
```
tdelabro, line 1065: you don't really need to create a new module here. Or if you do, it should be behind the feature flag. Here it will happen in the path, even if it is empty, when the feature is not enabled

tdelabro: I would recommend to make it a file. `zeroize.rs`. And put `mod` import behind the feature flag in felt/mod.rs

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -1062,6 +1062,18 @@ mod errors {
...
+    use super::*;
+
+    #[cfg(feature = "zeroing")]
+    impl zeroize::Zeroize for Felt {
```
tdelabro, line 1069: Can't you just use leverage https://docs.rs/zeroize/latest/zeroize/trait.DefaultIsZeroes.html?

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -1844,4 +1856,14 @@ mod test {
...
+
+    #[cfg(feature = "zeroing")]
+    #[test]
+    fn zeroing_felt() {
+        use zeroize::Zeroize;
+
+        let mut felt = Felt::from_hex_unchecked("0x01");
+        felt.zeroize();
+        assert_eq!(felt, Felt::ZERO);
+    }
```
tdelabro, line 1868: move this tests to the new file too

tdelabro (changes requested):

crates/starknet-types-core/src/felt/mod.rs
Trantorian1: Yep you're right. I was not aware of this trait, thanks for the info!

crates/starknet-types-core/src/felt/mod.rs
Trantorian1: Same as above, I feel like moving this to a separate file would be overkill? Wdyt?

crates/starknet-types-core/src/felt/mod.rs
tdelabro: felt/mod.rs is a beefy 2k lines of code. It's already painful to navigate it

tdelabro: But it's just a matter of taste of preference. So I won't force mine on the codebase

tdelabro (approved):

tdelabro:
@pefontana @dan-starkware @maciejka @0xLucqs can we get a second review?

crates/starknet-types-core/Cargo.toml
```diff
@@ -24,18 +24,20 @@ serde = { version = "1", optional = true, default-features = false, features = [
 lambdaworks-crypto = { version = "0.10.0", default-features = false, optional = true }
 parity-scale-codec = { version = "3.6", default-features = false, optional = true }
 lazy_static = { version = "1.5", default-features = false, optional = true }
+zeroize = { version = "1.8.1", default-features = false, optional = true }
 
 [features]
-default = ["std", "serde", "curve", "num-traits"]
+default = ["std", "serde", "curve", "num-traits", "zeroize"]
```
0xLucqs, line 30: are we sure we want that by default ?

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -1057,6 +1057,13 @@ mod errors {
     }
 }
 
+#[cfg(feature = "zeroize")]
+mod zeroing {
+    use super::*;
```
0xLucqs, line 1062: do you need anything else than `Felt`?

0xLucqs (changes requested):

crates/starknet-types-core/Cargo.toml
tdelabro: indeed, maybe not

Trantorian1:
Applied your suggestions @0xLucqs, thanks catching that!

0xLucqs (approved):

PR #101, opened 2024-10-31
test: use Vec::with_capacity avoid realloc mem

nkysg:
# Pull Request type
test: use Vec::with_capacity avoid realloc mem

- Bugfix
- Feature
- Code style update (formatting, renaming)
- Build-related changes
- Documentation content changes
- Testing
- Other (please describe):

Resolves: #NA

## What is the new behavior?

-
-
-

### DIFF
--- a/crates/starknet-types-core/src/felt/mod.rs
+++ b/crates/starknet-types-core/src/felt/mod.rs
@@ -1076,10 +1076,11 @@ mod test {
     // Helper function to generate a vector of bits for testing purposes
     fn generate_be_bits(x: &Felt) -> Vec<bool> {
         // Initialize an empty vector to store the expected bits
-        let mut bits = Vec::with_capacity(x.0.representative().limbs.len() * 8 * 8);
+        let representative = x.0.representative();
+        let mut bits = Vec::with_capacity(limbs.len() * 8 * 8);
 
         // Iterate over each limb in the representative of x
-        for limb in x.0.representative().limbs {
+        for limb in bits.limbs {
             // Convert the limb to a sequence of 8 bytes (u8) in big-endian
             let bytes = limb.to_be_bytes();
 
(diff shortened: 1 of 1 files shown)


crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -1081,7 +1081,7 @@ mod test {
     // Helper function to generate a vector of bits for testing purposes
     fn generate_be_bits(x: &Felt) -> Vec<bool> {
         // Initialize an empty vector to store the expected bits
-        let mut bits = Vec::new();
+        let mut bits = Vec::with_capacity(x.0.representative().limbs.len() * 8 * 8);
```
tdelabro, line 1084: The call to representative is not free at all.
Now there are two of those.

If you want to do this you should 

```
let representative = x.0.representative();

let mut bits = Vec::with_capacity(limbs.len() * 8 * 8);

for limb in representative.limbs {
```

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -1836,7 +1836,7 @@ mod test {
                 bytes[31 - (i >> 1)] |= 15 << (4 * (i & 1));
             }
             let h = Felt::from_bytes_be(&bytes);
-            let mut res = Vec::new();
+            let mut res = Vec::with_capacity(enc_len(n_nibbles));
```
tdelabro, line 1839: This is the test. No need to to optimize it. Especially when guessing the len require computation

tdelabro (changes requested):

crates/starknet-types-core/src/felt/mod.rs
tdelabro: This is not what you did.

tdelabro:
@nkysg run the tests locally before you push

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -1076,10 +1076,11 @@ mod test {
     // Helper function to generate a vector of bits for testing purposes
     fn generate_be_bits(x: &Felt) -> Vec<bool> {
         // Initialize an empty vector to store the expected bits
-        let mut bits = Vec::new();
+        let representative = x.0.representative();
+        let mut bits = Vec::with_capacity(limbs.len() * 8 * 8);
 
         // Iterate over each limb in the representative of x
-        for limb in x.0.representative().limbs {
+        for limb in bits.limbs {
```
tdelabro, line 1083: won't compile

tdelabro (changes requested):

tdelabro:
inactive

PR #108, opened 2024-12-25
feat: add checked add/mul/sub

tdelabro:

Checked operation traits are not available for Felt.
## What is the new behavior?

Following traits are implemented 
- CheckedAdd
- CheckedMul
- CheckedSub

No

## Other information

It would make sense to impl CheckedDiv too. But the trait requires that `Div` be implemented. Something we chose to not do so far. It may be worth reconsidering this.

### DIFF
--- a/crates/starknet-types-core/src/felt/num_traits_impl.rs
+++ b/crates/starknet-types-core/src/felt/num_traits_impl.rs
@@ -1,6 +1,8 @@
 use super::Felt;
 use num_bigint::{ToBigInt, ToBigUint};
-use num_traits::{FromPrimitive, Inv, One, Pow, ToPrimitive, Zero};
+use num_traits::{
+    CheckedAdd, CheckedMul, CheckedSub, FromPrimitive, Inv, One, Pow, ToPrimitive, Zero,
+};
 
 impl ToBigInt for Felt {
     /// Converts the value of `self` to a [`BigInt`].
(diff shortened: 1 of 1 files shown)


0xLucqs (approved):
against my will

crates/starknet-types-core/src/felt/num_traits_impl.rs
```diff
@@ -121,6 +123,39 @@ impl Pow<u128> for Felt {
...
+        } else {
+            Some(res)
+        }
+    }
+}
+
+impl CheckedMul for Felt {
+    fn checked_mul(&self, v: &Self) -> Option<Self> {
+        let res = self * v;
+        if res < *self {
```
cchudant, line 140: I believe that implementation is wrong,
Felt::TWO * Felt::MAX will probably not work the same way as `2u32.wrapping_add(u32::MAX)` returns 1

cchudant (changes requested):

crates/starknet-types-core/src/felt/num_traits_impl.rs
tdelabro: You are right. It should check against `max(self, v)`

I fixed it and added a test

tdelabro: but now there is a bug for mul by zero

tdelabro: fixed

cchudant (changes requested):
I do not believe this is a good enough check for the CheckedMul.
Here is another case:
```rs
println!("{:?}", Felt::from(u128::MAX).checked_mul(&Felt::from(u128::MAX)));
```
This I believe overflows, but does not return None.

Checking for overflow on the multiply operation after the fact is tricky. I don't know any way to do it if we don't have the divide operation.

Issue #107, opened 2024-12-17
dev: improve poseidon API


tdelabro:
crates/starknet-types-core/src/hash/poseidon.rs

So the poseidon hash is defined this [way](https://docs.starknet.io/architecture-and-concepts/cryptography/hash-functions/#poseidon_hash).

And ours is this way:
```
impl StarkHash for Poseidon {
    /// Computes the Poseidon hash of two Felts, as defined
    /// in <https://docs.starknet.io/documentation/architecture_and_concepts/Hashing/hash-functions/#poseidon_hash.>
    fn hash(felt_0: &Felt, felt_1: &Felt) -> Felt {
        Felt(PoseidonCairoStark252::hash(&felt_0.0, &felt_1.0))
    }
    /// Computes the Poseidon hash of an array of Felts, as defined
    /// in <https://docs.starknet.io/documentation/architecture_and_concepts/Cryptography/hash-functions/#poseidon_array_hash.>
    fn hash_array(felts: &[Felt]) -> Felt {
        // Non-copy but less dangerous than transmute
        // https://doc.rust-lang.org/std/mem/fn.transmute.html#alternatives
        Felt(PoseidonCairoStark252::hash_many(unsafe {
            core::slice::from_raw_parts(
                felts.as_ptr() as *const FieldElement<Stark252PrimeField>,
                felts.len(),
            )
        }))
    }
}

impl Poseidon {
    /// Computes the Hades permutation over a mutable state of 3 Felts, as defined
    /// in <https://docs.starknet.io/documentation/architecture_and_concepts/Cryptography/hash-functions/#poseidon_array_hash>
    pub fn hades_permutation(state: &mut [Felt; 3]) {
        let mut state_inner = [state[0].0, state[1].0, state[2].0];
        PoseidonCairoStark252::hades_permutation(&mut state_inner);
        for i in 0..3 {
            state[i] = Felt(state_inner[i]);
        }
    }
}
```

This means that if you want to hash many elements you use `hash_array`,  two elements you use `hash`. and for 1 single element you have to understand by yourselves that, according to poseidon specification, you must call `hades_permutation(my_value, 0, 1)`.
As you can probably guess people don't know that and just do crazy things instead.

The inner labmdawork API is way more useful for developers: https://github.com/lambdaclass/lambdaworks/blob/7b5a638d4ce81f380ea5f43a22be41ef9b2d7ff2/crypto/src/hash/poseidon/mod.rs#L17

[...]

PR #109, opened 2024-12-30
feat: improve poseidon API

varun-doshi:
# Pull Request type

StarkHash trait does have `hash_single` funciton

Resolves: https://github.com/starknet-io/types-rs/issues/107

## What is the new behavior?

- StarkHash now implements `hash_single` function

No

## Other information

This results in Pedersen needing to implement `hash_single` as well. Currently, it is set as todo!

### DIFF
--- a/crates/starknet-types-core/src/hash/pedersen.rs
+++ b/crates/starknet-types-core/src/hash/pedersen.rs
@@ -26,6 +26,10 @@ impl StarkHash for Pedersen {
             });
         Felt(PedersenStarkCurve::hash(&current_hash, &data_len.0))
     }
+    
+    fn hash_single(felt: &Felt) -> Felt {
+        todo!()
+    }
 }
 
 #[cfg(test)]
--- a/crates/starknet-types-core/src/hash/poseidon.rs
+++ b/crates/starknet-types-core/src/hash/poseidon.rs
@@ -28,6 +28,10 @@ impl StarkHash for Poseidon {
             )
         }))
     }
+
+    fn hash_single(felt: &Felt) -> Felt {
+        Felt(PoseidonCairoStark252::hash_single(&felt.0))
+    }
 }
 
 impl Poseidon {
@@ -45,6 +49,17 @@ impl Poseidon {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn test_poseidon_single() {
+        let x = Felt::from_hex("0x9").unwrap();
+
+        assert_eq!(
+            Poseidon::hash_single(&x),
+            Felt::from_hex("0x3bb3b91c714cb47003947f36dadc98326176963c434cd0a10320b8146c948b3")
+                .unwrap()
+        );
+    }
 
     #[test]
     fn test_poseidon_hash() {
--- a/crates/starknet-types-core/src/hash/traits.rs
+++ b/crates/starknet-types-core/src/hash/traits.rs
@@ -7,4 +7,6 @@ pub trait StarkHash {
     /// Computes the hash of an array of Felts,
     /// as defined in <https://docs.starknet.io/documentation/architecture_and_concepts/Hashing/hash-functions/#array_hashing.>
     fn hash_array(felts: &[Felt]) -> Felt;
+
+    fn hash_single(felt: &Felt) -> Felt;
 }


tdelabro:
@varun-doshi I think the pedersen hash_single impl should be `pedersen(0, value)`

tdelabro:
@varun-doshi let's go with `pedersen(value, 0)` (based on @liorgold2 feedbacks) and make it very explicit in the documentation that there is room for collision as:
`pedersen(value)` and `pedersen(value, 0)` will return the same values
and
`pedersen_array([value])` and `pedersen(pedersen(0, value), 1)` will too

Add the first example to the `hash_single` doc and the second one to the `hash_array` doc

crates/starknet-types-core/src/hash/pedersen.rs
```diff
@@ -26,12 +29,32 @@ impl StarkHash for Pedersen {
             });
         Felt(PedersenStarkCurve::hash(&current_hash, &data_len.0))
     }
+
+    /// Computes the Pedersen hash of a single Felt
+    ///
+    /// Warning: there is room for collision as:
+    /// pedersen(value) and pedersen(value, 0) will return the same values
```
tdelabro, line 36: /// Pedersen::hash_singlevalue) and Pedersen::hash(value, 0) will return the same values

crates/starknet-types-core/src/hash/pedersen.rs
```diff
@@ -17,6 +17,9 @@ impl StarkHash for Pedersen {
 
     /// Computes the Pedersen hash of an array of Felts, as defined
     /// in <https://docs.starknet.io/documentation/architecture_and_concepts/Hashing/hash-functions/#array_hashing.>
+    ///
+    /// Warning: there is room for collision as:
+    /// pedersen_array([value]) and pedersen(pedersen(0, value), 1) will return the same values
```
tdelabro, line 22: /// Pedersen::hash_array([value]) and Pedersen::hash(Pedersen::hash(0, value), 1) will return the same values

tdelabro (changes requested):

crates/starknet-types-core/src/hash/pedersen.rs
```diff
@@ -26,12 +29,32 @@ impl StarkHash for Pedersen {
             });
         Felt(PedersenStarkCurve::hash(&current_hash, &data_len.0))
     }
+
+    /// Computes the Pedersen hash of a single Felt
+    ///
+    /// Warning: there is room for collision as:
+    /// Pedersen::hash_singlevalue) and Pedersen::hash(value, 0) will return the same values
```
tdelabro, line 36: Pedersen::hash_single(value)

forgoten '('

tdelabro (changes requested):

tdelabro (approved):

crates/starknet-types-core/src/hash/pedersen.rs
```diff
@@ -26,12 +29,32 @@ impl StarkHash for Pedersen {
             });
         Felt(PedersenStarkCurve::hash(&current_hash, &data_len.0))
     }
+
+    /// Computes the Pedersen hash of a single Felt
+    ///
+    /// Warning: there is room for collision as:
+    /// Pedersen::hash_single(value) and Pedersen::hash(value, 0) will return the same values
```
jbcaron, line 36: Could you add a test that explicitly verifies the documented collision behavior between hash_single(value) and hash(value, 0)

jbcaron (approved):

tdelabro:
@varun-doshi it looks like your branch fails to build. Can you check what is going wrong?

varun-doshi:
> @varun-doshi it looks like your branch fails to build. Can you check what is going wrong?

@tdelabro it seems that command is running successfully on my local. This is build error showing up on CI
```
error: package `parity-scale-codec v3.7.1` cannot be built because it requires rustc 1.79.0 or newer, while the currently active rustc version is 1.77.2
Either upgrade to rustc 1.79.0 or newer, or use
cargo update parity-scale-codec@3.7.1 --precise ver
where `ver` is the latest version of `parity-scale-codec` supporting rustc 1.77.2
```

Is it possible its a CI issue? Pls correct me if im wrong

tdelabro:
@varun-doshi I will suggest you update https://github.com/starknet-io/types-rs/blob/main/rust-toolchain.toml
setting the toolchain to the latest rust release `1.84.1`

varun-doshi:
@tdelabro updating the toolchain version of this PR back to `1.80.1`(which is what is on main right now) after you confirm
Currently this PR is on `1.84.1` as you had suggested

jbcaron:
> @tdelabro updating the toolchain version of this PR back to `1.80.1`(which is what is on main right now) after you confirm Currently this PR is on `1.84.1` as you had suggested

The CI passing with version 1.80.1.

antiyro (approved):

PR #112, opened 2025-01-22
Implement addition and multiplication operations for AffinePoint

FilipLaurentiu:
- Feature

Addition or multiplication not allowed

Resolves: #NA

## What is the new behavior?
AffinePoints could now be added or multiplied by an exponent

No

### DIFF
--- a/crates/starknet-types-core/src/curve/affine_point.rs
+++ b/crates/starknet-types-core/src/curve/affine_point.rs
@@ -1,9 +1,14 @@
-use crate::curve::curve_errors::CurveError;
-use crate::felt::Felt;
-use lambdaworks_math::cyclic_group::IsGroup;
-use lambdaworks_math::elliptic_curve::short_weierstrass::curves::stark_curve::StarkCurve;
-use lambdaworks_math::elliptic_curve::short_weierstrass::point::ShortWeierstrassProjectivePoint;
-use lambdaworks_math::elliptic_curve::traits::FromAffine;
+use crate::{curve::curve_errors::CurveError, felt::Felt};
+use lambdaworks_math::{
+    cyclic_group::IsGroup,
+    elliptic_curve::{
+        short_weierstrass::{
+            curves::stark_curve::StarkCurve, point::ShortWeierstrassProjectivePoint,
+        },
+        traits::{FromAffine, IsEllipticCurve},
+    },
+    unsigned_integer::traits::IsUnsignedInteger,
+};
 
 /// Represents a point on the Stark elliptic curve.
 /// Doc: https://docs.starkware.co/starkex/crypto/stark-curve.html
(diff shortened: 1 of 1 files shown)


crates/starknet-types-core/src/curve/affine_point.rs
```diff
@@ -47,6 +51,15 @@ impl AffinePoint {
     pub fn y(&self) -> Felt {
         Felt(*self.0.y())
     }
+
+    pub fn generator() -> Self {
```
tdelabro, line 55: add documentation

crates/starknet-types-core/src/curve/affine_point.rs
```diff
@@ -47,6 +51,15 @@ impl AffinePoint {
...
+        AffinePoint(StarkCurve::generator())
+    }
+
+    pub fn mul(&self, exponent: Felt) -> Self {
```
tdelabro, line 59: add documentation

crates/starknet-types-core/src/felt/mod.rs
```diff
@@ -295,7 +295,7 @@ impl Felt {
 
     /// Quotient and remainder between `self` and `rhs`.
     pub fn div_rem(&self, rhs: &NonZeroFelt) -> (Self, Self) {
-        let (q, r) = self.0.representative().div_rem(&rhs.0.representative());
+        let (q, r) = self.representative().div_rem(&rhs.0.representative());
```
tdelabro, line 298: doesn't this change the behaviour?

tdelabro (changes requested):
Can you explain to me why is this something we want? Are are you sure there is no reason we could not want it to be implemented?

(I have no opinion, just trying to understand)

FilipLaurentiu:
> Can you explain to me why is this something we want? Are are you sure there is no reason we could not want it to be implemented?
> 
> (I have no opinion, just trying to understand)

`starknet-rs` library use this crate for their EC points representation. I am using `starknet-rs` for my project and I need these operations. The other option is to use the underlying `lambdaworks_math` library but this will add a new layer of abstraction and will make the code harder to understand. These are basic curve operation that are frequently used.

Another missing functionality is the ability to uncompress EC points that are used in Cairo

crates/starknet-types-core/src/felt/mod.rs
FilipLaurentiu: the `self.representative()` function returns `self.0.representative()` so it's the same. 
I had to add the public function `representative` to get the representation of a `Felt` for the `operate_with_self` function in order to multiply the point with itself.
Adding `representative()` function to `Felt` and replace the old way of using it make the code cleaner.

FilipLaurentiu:
Close this PR and open a new one with minimum changes and similar behavior as ProjectivePoint https://github.com/FilipLaurentiu/types-rs/pull/1

PR #113, opened 2025-01-23
chore: nuke starknet-types-rpc crate

tdelabro:
Nobody is using this crate. We are nuking it and abandoning support for it.


antiyro:
That's unfortunate and I still believe we should have proper maintainers on this repo, but if this is approved then we'll make sure to add clear native support in Madara. Thanks @tdelabro for handling this.

antiyro:
We should make sure this doesn't break any projects, but anyway, they would need to manifest themselves somewhere, like we did for Madara with the upcoming v0.8.0

Capture d’écran 2025-01-24 à 15 53 54

0xLucqs (approved):
kaboom

jbcaron (approved):

PR #127, opened 2025-04-11
Create new AffinePoint from the compressed point representation

FilipLaurentiu:
Cairo represent points on the curve just with the x coordinate (compressed). This is enough for signature verification but not enough for other operations on EC points.
This update provide a way to create a new `AffinePoint` from the x cordinate and the parity bit.

# Pull Request type

- Feature

- No option to create a new AffinePoint instance from the compressed representation.

## What is the new behavior?

- Create AffinePoints just from the x compressed representation

No


tdelabro (approved):

crates/starknet-types-core/src/curve/affine_point.rs
```diff
@@ -33,6 +34,28 @@ impl AffinePoint {
         ]))
     }
 
+    /// Construct new affine point from the `x` coordinate and the parity bit.
+    /// If the parity bit is false, choose the y-coordinate with even parity.
+    /// If the parity bit is true, choose the y-coordinate with odd parity.
```
gabrielbosio, line 39: For clarity:

```suggestion
    /// Construct new affine point from the `x` coordinate and the parity bit `y_parity`.
    /// If `y_parity` is false, choose the y-coordinate with even parity.
    /// If `y_parity` is true, choose the y-coordinate with odd parity.
```

gabrielbosio (changes requested):
Looks good, I have left a small comment on the function documentation

gabrielbosio (approved):

tdelabro:
Thanks @FilipLaurentiu, sorry for the 4 month it took us to merge this. 
We will be more active on this repo from now on

FilipLaurentiu:
> Thanks @FilipLaurentiu, sorry for the 4 month it took us to merge this. We will be more active on this repo from now on

Hey ! Thanks for the review and for the approval :+1: 

About the delay, the changes were not so impactful, but good to have, avoid the confusion about the point compression, but I notice delays on other, more important repos. 
Starknet will benefit from giving more attention to these repos :pray:

PR #139, opened 2025-07-10
feat: add cairo short string logic

### LINKED ISSUE #122: feat: add short string support
Felt can be interpreted as short string.

It would be nice to have a way to do
Felt -> ShortString
ShortString -> Felt
a macro to create Felt from 'static str

tdelabro:
No breaking change.
Only add features.

closes #122

### DIFF
--- a/crates/starknet-types-core/src/short_string/mod.rs
+++ b/crates/starknet-types-core/src/short_string/mod.rs
@@ -9,6 +9,7 @@
 //!
 //! The convesion to `Felt` is done by using the internal ascii short string as bytes and parse those as a big endian number.
 
+#[cfg(not(feature = "std"))]
 use crate::felt::alloc::string::{String, ToString};
 use crate::felt::Felt;
 
@@ -114,10 +115,7 @@ impl TryFrom<&str> for ShortString {
 
 #[cfg(test)]
 mod tests {
-    use crate::{
-        felt::Felt,
-        short_string::{ShortString, TryShortStringFromStringError},
-    };
+    use super::*;
 
     #[test]
     fn ok() {
--- a/crates/starknet-types-core/src/short_string/mod.rs
+++ b/crates/starknet-types-core/src/short_string/mod.rs
@@ -78,13 +78,13 @@ impl TryFrom<String> for ShortString {
 }
 
 impl Felt {
-    /// Create a felt value from a cairo short string
+    /// Create a felt value from a cairo short string.
     ///
     /// The string must contains only ascii characters
     /// and its length must not exceed 31.
-    pub fn try_from_cairo_short_string(
-        string: &str,
-    ) -> Result<Self, TryShortStringFromStringError> {
+    ///
+    /// The returned felt value be that of the input raw bytes.
+    pub fn parse_cairo_short_string(string: &str) -> Result<Self, TryShortStringFromStringError> {
         let bytes = string.as_bytes();
         if !bytes.is_ascii() {
             return Err(TryShortStringFromStringError::NonAscii);
@@ -139,7 +139,7 @@ mod tests {
                 ]),
             ),
         ] {
-            let felt = Felt::try_from_cairo_short_string(&string).unwrap();
+            let felt = Felt::parse_cairo_short_string(&string).unwrap();
             let short_string = ShortString::try_from(string.clone()).unwrap();
 
             assert_eq!(felt, expected_felt);
@@ -151,13 +151,13 @@ mod tests {
     #[test]
     fn ko_too_long() {
         let ok_string = String::from("This is a 31 characters string.");
-        assert!(Felt::try_from_cairo_short_string(&ok_string).is_ok());
+        assert!(Felt::parse_cairo_short_string(&ok_string).is_ok());
         assert!(ShortString::try_from(ok_string).is_ok());
 
         let ko_string = String::from("This is a 32 characters string..");
 
         assert_eq!(
-            Felt::try_from_cairo_short_string(&ko_string),
+            Felt::parse_cairo_short_string(&ko_string),
             Err(TryShortStringFromStringError::TooLong)
         );
         assert_eq!(
@@ -171,7 +171,7 @@ mod tests {
         let string = String::from("What a nice emoji 💫");
 
         assert_eq!(
-            Felt::try_from_cairo_short_string(&string),
+            Felt::parse_cairo_short_string(&string),
             Err(TryShortStringFromStringError::NonAscii)
         );
         assert_eq!(
--- a/crates/starknet-types-core/src/short_string/mod.rs
+++ b/crates/starknet-types-core/src/short_string/mod.rs
@@ -1,11 +1,21 @@
+//! Cairo short string
+//!
+//! The cairo language make it possible to create `Felt` values at compile time from so-called "short string".
+//! See https://docs.starknet.io/archive/cairo-101/strings/ for more information and syntax.
+//!
+//! This modules allows to mirror this behaviour in Rust, by leveraging type safety.
+//! A `ShortString` is string that have been checked and is guaranted to be convertible into a valid `Felt`.
+//! It checks that the `String` only contains ascii characters and is no longer than 31 characters.
+//!
+//! The convesion to `Felt` is done by using the internal ascii short string as bytes and parse those as a big endian number.
+
+use crate::felt::alloc::string::{String, ToString};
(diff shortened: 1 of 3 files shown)


maciejka (approved):

crates/starknet-types-core/src/short_string/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+        self.0.fmt(f)
+    }
+}
+
+impl From<ShortString> for Felt {
+    fn from(ss: ShortString) -> Self {
+        let bytes = ss.0.as_bytes();
+
+        let mut buffer = [0u8; 32];
+        buffer[(32 - bytes.len())..].copy_from_slice(bytes);
```
0xLucqs, line 35: if `bytes.len() > 32` you're fucked

crates/starknet-types-core/src/short_string/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+        if !bytes.is_ascii() {
+            return Err(TryShortStringFromStringError::NonAscii);
+        }
+
+        let mut buffer = [0u8; 32];
+        buffer[(32 - bytes.len())..].copy_from_slice(bytes);
+
+        // The conversion will never fail
+        Ok(Felt::from_bytes_be(&buffer))
+    }
```
0xLucqs, line 98: why not using the `TryFrom` impl ? and naming it `try_from_cairo_short_string` is confusing imo as you're expecting a `&str`

crates/starknet-types-core/src/short_string/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+            let short_string = ShortString::try_from(string.clone()).unwrap();
+
+            assert_eq!(felt, expected_felt);
+            assert_eq!(short_string.0, string);
+            assert_eq!(Felt::from(short_string), expected_felt);
+        }
+    }
+
+    #[test]
+    fn ko_too_long() {
```
0xLucqs, line 149: test the too long case for the `From` impl

crates/starknet-types-core/src/short_string/mod.rs
tdelabro: `try_from_cairo_short_string` you find it confusing because it takes a `&str` but the name says `string`.

We are not referring to rust `String` type here but to "cairo short string?, which is another concept, that just happen to also use the the word "string". Not my choice of naming, but that is what those are called.
Mentioning that those are "cairo short" help for the disambiguation tho

crates/starknet-types-core/src/short_string/mod.rs
tdelabro: Due to the way `ShortString` is initialized, see `impl TryFrom<String> for ShortString`, there is no way `bytes.len()` can be greater than 32.

The string is guaranteed to be `31` chars at most, and the `is_ascii` guarantee that each one of those characters is made of 1 byte. Therefore, `bytes.len()` cannot be more than `31`

AvivYossef-starkware (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starknet-io/types-rs/139)* status: all files reviewed, 4 unresolved discussions (waiting on @0xLucqs, @dan-starkware, and @pefontana)

___
*[`crates/starknet-types-core/src/short_string/mod.rs` line 57 at r1](https://reviewable.io/reviews/starknet-io/types-rs/139#-OV6np0-AWxEB6AUdY2l:-OV6np001VXgi--nHsHM:bbjq41u) ([raw file](https://github.com/starknet-io/types-rs/blob/b6b01ef6d4bbc7c8e8f8e42213ae864818fa5d51/crates/starknet-types-core/src/short_string/mod.rs#L57)):*
> ```XML
>         .fmt(f)
>     }
> }
> ```

use thiserror instead
https://docs.rs/thiserror/latest/thiserror/



_Code quote:_
```XML
#[derive(Debug, Clone)]
#[cfg_attr(test, derive(PartialEq, Eq))]
pub enum TryShortStringFromStringError {
    TooLong,
    NonAscii,
}

impl core::fmt::Display for TryShortStringFromStringError {
    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
        match self {
            TryShortStringFromStringError::TooLong => "string to long",
            TryShortStringFromStringError::NonAscii => "string contains non ascii characters",
        }
        .fmt(f)
    }
}
```
___
*[`crates/starknet-types-core/src/lib.rs` line 9 at r1](https://reviewable.io/reviews/starknet-io/types-rs/139#-OV6mxlV5BXY9b6NMqxH:-OV6mxlV5BXY9b6NMqxI:b-csp0df) ([raw file](https://github.com/starknet-io/types-rs/blob/b6b01ef6d4bbc7c8e8f8e42213ae864818fa5d51/crates/starknet-types-core/src/lib.rs#L9)):*
> ```XML
> pub mod felt;
> 
> #[cfg(any(feature = "std", feature = "alloc"))]
> ```

Why under the feature?

crates/starknet-types-core/src/short_string/mod.rs
gabrielbosio: The following snippet:

```rust
let ko_string = String::from("This is a 33 characters string...");
let _ = Felt::from(ShortString(ko_string));
```

fails with this error:

```
thread [...] panicked at crates/starknet-types-core/src/short_string/mod.rs:36:16:
attempt to subtract with overflow
```

If the idea is to panic when the string is too long then I think it would be great to have a more descriptive error message

gabrielbosio: Now that I realize, I tested it by adding a test case in the same file so I could create a `ShortString` using the inner value.
So I think that, instead of improving the error message (because that from function would never panic), it would be good to have a comment that explains why the subtraction would never underflow.

crates/starknet-types-core/src/short_string/mod.rs
0xLucqs: it's confusing because of what gabriel explained. It's called from cairo short string but doesn't expect a short string. Imo this should be `impl From<ShortString> for Felt`

crates/starknet-types-core/src/short_string/mod.rs
tdelabro: @gabrielbosio yup the inner value is only accessible from this mod. Outside of it your only way to create one is to go through this impl `impl TryFrom<&str> for ShortString`

> why the subtraction would never underflow
I can add a comment, giving some context for sure

crates/starknet-types-core/src/short_string/mod.rs
tdelabro: @0xLucqs `impl From<ShortString> for Felt` do already exists. But that means that you will have to manually do:
```rust
let sn_sepolia_id = Felt::from(ShortString::try_from("SN_SEPOLIA")?);
```
which is verbose and requires understanding the impl detail that `ShortString` is.

This function is a handy shortcut for developers that want to parse a cairo short string.
```rust
let sn_sepolia_id = Felt::try_from_cairo_short_string("SN_SEPOLIA")?;
```

It's expressive, it's straightforward, and it is even a bit more optimized as it avoids the allocation of a new string, which comes with the initialization of a `ShortString`

tdelabro: @gabrielbosio 

> One should expect to pass a ShortString as an argument to this function, not a &str.

Yeah I get your point.

> /// Create a felt value from a string. This function converts the passed string to a `ShortString` and then it converts the result to a `Felt`, returning it.

The only issue is that it is not what the implementation does. It doesn't internally create a `ShortString` type, because it doesn't need to.

0xLucqs: `short_string!("some short string");`

tdelabro: This could be a great addition, to execute at compile time and crash if the provided value is not valid.
But we also should provide a solution for runtime parsing of `String` (`parse_cairo_short_string(&a_string)`)

tdelabro: No, we don't want to use the intermediate `ShortString` type there. It is suboptimal because it requires creating `String`, which will involve allocation if input is a `&str`.
Those are two slightly different use cases.

Also, it is way less user-friendly and expressive than `Felt::parse_cairo_short_string(&a_string)` which describes exactly what it does

PR #144, opened 2025-07-26
Filip/secret felt implementation

FilipLaurentiu:
# Pull Request type

- Feature

`Felt` implement the Copy trait and it could leak the secret information into memory.
`Felt` is not zeroized on drop, it need to be manually zeroized.

Resolves: #NA

## What is the new behavior?

New type `SecretFelt` is introduced for this specific purpose. This type could be used to represent sensible information that will not be leaked into memory at the end of the program

No - Only if a project use the `zeroize` function on `Felt` and upgrade to the latest version of this library, but from what I know, nobody use it (starknet-rs doesn't use it)


tdelabro:
> No - Only if a project use the zeroize function on Felt and upgrade to the latest version of this library, but from what I know, nobody use it (starknet-rs doesn't use it)

So yes haha ;)

crates/starknet-types-core/src/felt/secret_felt.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// # Example
+    ///
+    /// ```
+    /// use starknet_types_core::felt::{Felt, secret_felt::SecretFelt};
+    ///
+    /// let mut private_key = Felt::from_hex_unchecked("0x123...");
+    /// let secret_felt = SecretFelt::from_felt(&mut private_key);
+    /// // private_key is now zeroized (set to Felt::ZERO)
+    /// ```
+    pub fn from_felt(secret_scalar: &mut Felt) -> Self {
```
tdelabro, line 40: maybe add a warning saying that this is only safe if the function is called immediately after the Felt was created. Because if the value is moved around before this call, it may have left some copy of its value here and there in memory

crates/starknet-types-core/src/felt/secret_felt.rs
```diff
@@ -0,0 +1,201 @@
...
+        Ok(secret_felt)
+    }
+
+    /// Provides reference access to the secret scalar.
+    ///
+    /// # Warning
+    ///
+    /// Be careful not to copy the value elsewhere, as that would defeat
+    /// the security guarantees of this type.
+    pub fn secret_scalar(&self) -> &Felt {
```
tdelabro, line 95: Is there a way to prevent this? Return a non-copy wrapper around a felt?

tdelabro (changes requested):

crates/starknet-types-core/src/felt/secret_felt.rs
FilipLaurentiu: I think I found a good solution. Now it returnes a `Zeroizing<Felt>` which is safe, it will be zeroized on drop, without the manual `.zeroize()` call

0xLucqs (approved):

tdelabro:
@FilipLaurentiu CI is failing

crates/starknet-types-core/src/felt/secret_felt.rs
```diff
@@ -0,0 +1,233 @@
...
+    /// let mut private_key = String::from_utf8(vec![255,255,..]).unwrap();
+    /// let secret_felt = SecretFelt::from_hex_string(&mut private_key);
+    /// ```
+    pub fn from_hex_string(hex: &mut String) -> Result<Self, FromStrError> {
+        let secret_felt = Felt::from_hex(&hex)?;
+        hex.zeroize();
+        Ok(Self(Box::new(secret_felt)))
+    }
+
+    /// Creates a new [SecretFelt] from its big-endian representation in a Vec<u8> of length 32.
```
tdelabro, line 68: mention that the input will be zeroized as a side effect

crates/starknet-types-core/src/felt/secret_felt.rs
```diff
@@ -0,0 +1,233 @@
...
+            .try_into()
+            .map_err(|_| ByteConversionError::InvalidValue)?;
+        let secret_felt = Self(Box::new(Felt::from_bytes_be(&value)));
+        secret.zeroize();
+        value.zeroize();
+        Ok(secret_felt)
+    }
+
+    /// Creates a new [SecretFelt] from its little-endian representation in a Vec<u8> of length 32.
+    /// Internally it uses [from_bytes_le](Felt::from_bytes_le).
```
tdelabro, line 82: mention that the input will be zeroized as a side effect

crates/starknet-types-core/src/felt/secret_felt.rs
```diff
@@ -0,0 +1,233 @@
...
+    ///
+    /// # Warning
+    ///
+    /// Be careful not to copy the value elsewhere, as that would defeat
+    /// the security guarantees of this type.
+    pub fn inner_value(&self) -> Zeroizing<Felt> {
+        Zeroizing::new(*self.0.clone())
+    }
+}
+
```
tdelabro, line 104: Should we impl `asRef<Felt>` and `AsRefMut<Felt>`?
Those are accessible through `inner_value` as they are implemented on `Zeroizing`

tdelabro (changes requested):

crates/starknet-types-core/src/felt/secret_felt.rs
tdelabro: okay great!

crates/starknet-types-core/src/felt/secret_felt.rs
```diff
@@ -2,6 +2,9 @@ use crate::felt::{Felt, FromStrError};
 use lambdaworks_math::errors::ByteConversionError;
 use zeroize::{Zeroize, Zeroizing};
 
+#[cfg(feature = "alloc")]
```
tdelabro, line 5: This will always be true due to your feature flag configuration:

```
secret_felt = ["alloc", "dep:zeroize"]
```

for this file to be compiled `secret_felt` will be enabled and it now require `alloc`.

So you can just remove this cfg marco altogether

tdelabro (changes requested):

crates/starknet-types-core/src/felt/secret_felt.rs
FilipLaurentiu: The intention was to use the imports only when `no_std` was needed.
I think I got it right this time.

crates/starknet-types-core/src/felt/secret_felt.rs
```diff
@@ -0,0 +1,280 @@
...
+    ///
+    /// Be careful not to copy the value elsewhere, as that would defeat
+    /// the security guarantees of this type.
+    pub fn inner_value(&self) -> Zeroizing<Felt> {
+        Zeroizing::new(*self.0.clone())
+    }
+}
+
+/// Constant time equality check for [SecretFelt]
+impl PartialEq for SecretFelt {
```
tdelabro, line 127: I think we should also implement `Eq`.
The additional reflexivity requirement is fulfilled for `Felt`

https://doc.rust-lang.org/std/cmp/trait.Eq.html

Am I missing something?

tdelabro (changes requested):

crates/starknet-types-core/Cargo.toml
```diff
@@ -49,18 +52,18 @@ serde = ["alloc", "dep:serde"]
 prime-bigint = ["dep:lazy_static"]
 num-traits = []
 papyrus-serialization = ["std"]
-zeroize = ["dep:zeroize"]
+secret_felt = ["alloc", "dep:zeroize", "dep:subtle", "subtle/const-generics", "subtle/core_hint_black_box", "dep:rand", "rand/alloc"]
```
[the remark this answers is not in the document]
FilipLaurentiu: Done :+1:

tdelabro (approved):

PR #149, opened 2025-08-12
Add `QM31` implementation

FrancoGiachetta:
# Summary

The purpose of this PR is to move QM31's [implementation](https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/math_utils/mod.rs#L89) from Cairo VM to types-rs. It adds a wrapper over `FieldElement<Stark252PrimeField>` which allows a safer usage of QM31 operations. 
The implemented set of operations only corresponds to the ones in Cairo VM.

# Pull Request type

- Feature

QM31 is not implemented in types-rs

Resolves: #NA

## What is the new behavior?

Add `QM31Felt` abstraction for performing QM31 operation in a safe manner.

No

### DIFF
--- a/crates/starknet-types-core/src/felt/alloc_impls.rs
+++ b/crates/starknet-types-core/src/felt/alloc_impls.rs
@@ -21,12 +21,7 @@ impl Felt {
     /// The resulting string is guaranted to be 66 chars long, which is enough to represent `Felt::MAX`:
     /// 2 chars for the `0x` prefix and 64 chars for the padded hexadecimal felt value.
     pub fn to_fixed_hex_string(&self) -> alloc::string::String {
-        let hex_str = alloc::format!("{self:#x}");
-        if hex_str.len() < 66 {
-            alloc::format!("0x{:0>64}", hex_str.strip_prefix("0x").unwrap())
-        } else {
-            hex_str
-        }
+        alloc::format!("{self:#066x}")
     }
 }
 
@@ -35,22 +30,7 @@ impl fmt::LowerHex for Felt {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let hex = alloc::string::ToString::to_string(&self.0);
         let hex = hex.strip_prefix("0x").unwrap();
-
-        let width = if f.sign_aware_zero_pad() {
-            f.width().unwrap().min(64)
-        } else {
-            1
-        };
-        if f.alternate() {
-            write!(f, "0x")?;
-        }
-
-        if hex.len() < width {
-            for _ in 0..(width - hex.len()) {
-                write!(f, "0")?;
-            }
-        }
-        write!(f, "{}", hex)
+        f.pad_integral(true, if f.alternate() { "0x" } else { "" }, hex)
     }
 }
 
@@ -59,22 +39,7 @@ impl fmt::UpperHex for Felt {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let hex = alloc::string::ToString::to_string(&self.0);
         let hex = hex.strip_prefix("0x").unwrap().to_uppercase();
-
-        let width = if f.sign_aware_zero_pad() {
-            f.width().unwrap().min(64)
-        } else {
-            1
-        };
-        if f.alternate() {
-            write!(f, "0x")?;
-        }
-
-        if hex.len() < width {
-            for _ in 0..(width - hex.len()) {
-                write!(f, "0")?;
-            }
-        }
-        write!(f, "{}", hex)
+        f.pad_integral(true, if f.alternate() { "0x" } else { "" }, &hex)
     }
 }
 


crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,447 @@
...
+                "Number used as QM31 since it's more than 144 bits long: {}",
+                felt
+            ),
+        }
+    }
+}
+
+#[repr(transparent)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct QM31Felt(pub(crate) FieldElement<Stark252PrimeField>);
```
JulianGCalderon, line 35: All `QM31` operations actually convert the `Felt` to the raw representation (4 coordinates), and operate with them directly.

What if we store the `QM31` in the raw representation directly (`[u64;4]`), and just provide the conversion to/from `Felt`?

Also, I'd keep the internal representation private so that we can later change how we represent it.

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,447 @@
...
+};
+
+pub const STWO_PRIME: u64 = (1 << 31) - 1;
+const STWO_PRIME_U128: u128 = STWO_PRIME as u128;
+const MASK_36: u64 = (1 << 36) - 1;
+const MASK_8: u64 = (1 << 8) - 1;
+
+#[derive(Debug)]
+pub enum QM31Error {
+    InvalidQM31(Felt),
```
JulianGCalderon, line 15: This error is only used when converting a `Felt` to a `QM31`, so I'd rename this variant to something like "FeltTooBig", with a message like "The felt is longer than 144 bits: {}".

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,447 @@
...
+    fn from(value: &QM31Felt) -> Self {
+        Felt(value.0)
+    }
+}
+
+impl From<QM31Felt> for Felt {
+    fn from(value: QM31Felt) -> Self {
+        Felt(value.0)
+    }
+}
```
JulianGCalderon, line 248: Are both implementations required? (idem for `TryFrom<Felt> for QM31`)

JulianGCalderon: If yes, is there a way to reuse the code? Maybe calling the `TryFrom<&Felt>` implementation in `TryFrom<Felt>`.

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,447 @@
...
+        result_bytes[0..9].copy_from_slice(&bytes_part1[0..9]);
+        result_bytes[9..18].copy_from_slice(&bytes_part2[0..9]);
+
+        let value = Felt::from_bytes_le(&result_bytes);
+
+        Self(value.0)
+    }
+
+    /// Computes the addition of two QM31 elements in reduced form.
+    /// Returns an error if either operand is not reduced.
```
JulianGCalderon, line 74: This comment indicates that the function returns an error, but it does not return a result.

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,447 @@
...
+            coordinates1[0] + coordinates2[0],
+            coordinates1[1] + coordinates2[1],
+            coordinates1[2] + coordinates2[2],
+            coordinates1[3] + coordinates2[3],
+        ];
+        Self::from_raw(result_unreduced_coordinates)
+    }
+
+    /// Computes the negative of a QM31 element in reduced form.
+    /// Returns an error if the input is not reduced.
```
JulianGCalderon, line 88: This comment indicates that the function returns an error, but it does not return a result.

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,447 @@
...
+        Self::from_raw([
+            STWO_PRIME - coordinates[0],
+            STWO_PRIME - coordinates[1],
+            STWO_PRIME - coordinates[2],
+            STWO_PRIME - coordinates[3],
+        ])
+    }
+
+    /// Computes the subtraction of two QM31 elements in reduced form.
+    /// Returns an error if either operand is not reduced.
```
JulianGCalderon, line 100: This comment indicates that the function returns an error, but it does not return a result.

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,447 @@
...
+            STWO_PRIME + coordinates1[0] - coordinates2[0],
+            STWO_PRIME + coordinates1[1] - coordinates2[1],
+            STWO_PRIME + coordinates1[2] - coordinates2[2],
+            STWO_PRIME + coordinates1[3] - coordinates2[3],
+        ];
+        Self::from_raw(result_unreduced_coordinates)
+    }
+
+    /// Computes the multiplication of two QM31 elements in reduced form.
+    /// Returns an error if either operand is not reduced.
```
JulianGCalderon, line 114: This comment indicates that the function returns an error, but it does not return a result.

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,447 @@
...
+                - coordinates[2] * denom_inverse_i
+                - coordinates[3] * denom_inverse_r,
+        ])
+    }
+
+    /// Computes the division of two QM31 elements in reduced form.
+    /// Returns an error if the input is zero.
+    /// # Safety
+    /// Will panic if the rhs value is equal to zero.
+    pub fn div(&self, rhs: &QM31Felt) -> Result<QM31Felt, QM31Error> {
```
JulianGCalderon, line 214: The signature of the function contains a result, but it will panic instead of returning an error. Can we remove the panic and keep the result?

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,447 @@
...
+                + coordinates1_u64[1] * coordinates2_u64[2]
+                + coordinates1_u64[2] * coordinates2_u64[1]
+                + coordinates1_u64[3] * coordinates2_u64[0],
+        ];
+        Self::from_raw(result_coordinates)
+    }
+
+    /// Computes the inverse in the M31 field using Fermat's little theorem, i.e., returns
+    /// `v^(STWO_PRIME-2) modulo STWO_PRIME`, which is the inverse of v unless v % STWO_PRIME == 0.
+    fn pow2147483645(v: u64) -> u64 {
```
JulianGCalderon, line 150: Given that this function calculates the inverse of a M31 field element. Maybe we should rename this function to `m31_inverse`?

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,447 @@
...
+    fn sqn(v: u64, n: usize) -> u64 {
+        let mut u = v;
+        for _ in 0..n {
+            u = (u * u) % STWO_PRIME;
+        }
+        u
+    }
+
+    /// Computes the inverse of a QM31 element in reduced form.
+    /// Returns an error if the denominator is zero or either operand is not reduced.
```
JulianGCalderon, line 170: This comment indicates that the function returns an error, but it panics instead.

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Changed in [21ec03e](https://github.com/starknet-io/types-rs/pull/149/commits/21ec03e5ecce18dd4ca2eb97e678e23825b18af7)

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Changed in [21ec03e](https://github.com/starknet-io/types-rs/pull/149/commits/21ec03e5ecce18dd4ca2eb97e678e23825b18af7)

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Changed in [21ec03e](https://github.com/starknet-io/types-rs/pull/149/commits/21ec03e5ecce18dd4ca2eb97e678e23825b18af7)

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    pub fn is_zero(&self) -> bool {
+        *self == Self::ZERO
+    }
+}
+
+impl From<&QM31Felt> for Felt {
+    fn from(value: &QM31Felt) -> Self {
+        Felt::from_raw(value.0)
```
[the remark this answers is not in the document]
JulianGCalderon: Also, we should add some tests on the conversion `Felt -> QM31 -> Felt`.

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,427 @@
...
+            felt.to_le_digits()
+        };
+
+        Self([
+            (limbs[0] & MASK_36),
+            ((limbs[0] >> 36) + ((limbs[1] & MASK_8) << 28)),
+            ((limbs[1] >> 8) & MASK_36),
+            ((limbs[1] >> 44) + (limbs[2] << 20)),
+        ])
+    }
```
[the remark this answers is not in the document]
FrancoGiachetta: The `from_raw` method reduces the given coordinates. Since the are reduced and represented as bytes, I first create a Felt so that I can retrieve the reduces coordinates. If they are not reduced, we can endup having `QM31Felt`s that represent more than 144 bits.

As for `as_raw` its a getter.

crates/starknet-types-core/src/felt/qm31.rs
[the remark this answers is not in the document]
FrancoGiachetta: Added a test for this here: [a7691d7](https://github.com/starknet-io/types-rs/pull/149/commits/a7691d78058ab01b6fd5e90c233490f7073298a2)

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,427 @@
...
+    pub fn add(&self, rhs: &QM31Felt) -> QM31Felt {
+        let coordinates1 = self.as_raw();
+        let coordinates2 = rhs.as_raw();
+        let result_unreduced_coordinates = [
+            coordinates1[0] + coordinates2[0],
+            coordinates1[1] + coordinates2[1],
+            coordinates1[2] + coordinates2[2],
+            coordinates1[3] + coordinates2[3],
+        ];
+        Self::from_raw(result_unreduced_coordinates)
```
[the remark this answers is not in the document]
FrancoGiachetta: We do need the `from_raw` this to keep the result in the field. If not, the operation could overflow.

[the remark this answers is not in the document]
FrancoGiachetta: The `as_raw` is just a getter a reuse to avoid using the raw `self.0`. But could be changed if you prefer.

crates/starknet-types-core/src/felt/qm31.rs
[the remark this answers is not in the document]
FrancoGiachetta: I don't understand why this is not correct.

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,487 @@
...
+        match self {
+            QM31Error::UnreducedFelt(felt) => writeln!(
+                f,
+                "Number is not a packing of a QM31 in reduced form: {felt})"
+            ),
+            QM31Error::FeltTooBig(felt) => writeln!(
+                f,
+                "Number used as QM31 since it's more than 144 bits long: {felt}"
+            ),
+            QM31Error::InvalidInversion => writeln!(f, "Attempt to invert a qm31 equal to zero"),
```
tdelabro, line 31: I don't think `writeln` should be used inside `fmt` impl.
It's the caller responsibility to decide if he want to add a newline.

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,487 @@
+use core::fmt;
+
+use crate::felt::Felt;
+
+pub const STWO_PRIME: u64 = (1 << 31) - 1;
+const STWO_PRIME_U128: u128 = STWO_PRIME as u128;
+const MASK_36: u64 = (1 << 36) - 1;
+const MASK_8: u64 = (1 << 8) - 1;
+
+#[derive(Debug)]
```
tdelabro, line 10: derive Copy, Clone

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,487 @@
...
+                "Number used as QM31 since it's more than 144 bits long: {felt}"
+            ),
+            QM31Error::InvalidInversion => writeln!(f, "Attempt to invert a qm31 equal to zero"),
+        }
+    }
+}
+
+#[repr(transparent)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct QM31Felt([u64; 4]);
```
tdelabro, line 38: Add doc

tdelabro: Should explain what is the inner repr as it has impact of ordering

tdelabro: right now I don't know if it is big or little endian. It is important because it dictates the ordering

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,487 @@
...
+
+impl QM31Felt {
+    /// [QM31Felt] constant that's equal to 0.
+    pub const ZERO: Self = Self([0, 0, 0, 0]);
+
+    pub fn as_raw(&self) -> [u64; 4] {
```
tdelabro, line 44: doc

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,487 @@
...
+        self.0
+    }
+
```
tdelabro, line 47: the doc is misleading:
`from the raw internal representation` means that the argument to this function will be the value used for internal representation. But actually there is some processing involved before storing it

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,487 @@
...
+    /// Create a [QM31Felt] from the raw internal representation. Reduces four u64 coordinates so that they fit in 144 bits.
+    pub fn from_raw(coordinates: [u64; 4]) -> QM31Felt {
+        Self([
+            coordinates[0] % STWO_PRIME,
+            coordinates[1] % STWO_PRIME,
+            coordinates[2] % STWO_PRIME,
+            coordinates[3] % STWO_PRIME,
```
tdelabro, line 54: This is a one-way destructive operation.
Naming it `from_raw`, given that there is an `as_raw` method, imply that 
`x == QM31Felt::from_raw(x.as_raw())`
Which is not always true. 
Should probably rename it `from_coordinates`

crates/starknet-types-core/src/felt/qm31.rs
tdelabro: maybe `inner` rather than `as_raw` since it is not the raw value.

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,487 @@
...
+        ])
+    }
+
+    /// Packs the [QM31Felt] coordinates into a Felt.
+    pub fn pack_into_felt(&self) -> Felt {
```
tdelabro, line 59: Do we want to expose this one as `pub` when we implement `From/Into` later on?
Is there a good reason to offer the user more than one way to do this?

crates/starknet-types-core/src/felt/qm31.rs
tdelabro: why did we chose this internal repr? It should be documentend?

tdelabro (changes requested):
Do you want to add the num-trait impl in a further PR?

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Why is it necessary to derive Copy and Clone in this struct?

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: All the operations a previous conversion to coordinates. To avoid having to do this, the internal representation are the coordinates them selves.

FrancoGiachetta: Added doc here: [5bca78c](https://github.com/starknet-io/types-rs/pull/149/commits/5bca78caf2ef77032db715a55e0a2ca06f668922)

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Changed the method's name to `inner` here: [5bca78c](https://github.com/starknet-io/types-rs/pull/149/commits/5bca78caf2ef77032db715a55e0a2ca06f668922)

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Yes, you are absolutely right. Change the name to `inner` as suggested here: [5bca78c](https://github.com/starknet-io/types-rs/pull/149/commits/5bca78caf2ef77032db715a55e0a2ca06f668922)

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,489 @@
...
+        match self {
+            QM31Error::UnreducedFelt(felt) => write!(
+                f,
+                "Number is not a packing of a QM31 in reduced form: {felt})"
+            ),
+            QM31Error::FeltTooBig(felt) => write!(
+                f,
+                "Number used as QM31 since it's more than 144 bits long: {felt}"
+            ),
+            QM31Error::InvalidInversion => write!(f, "Attempt to invert a qm31 equal to zero"),
```
tdelabro, line 31: https://rust-lang.github.io/api-guidelines/interoperability.html#error-types-are-meaningful-and-well-behaved-c-good-err
```
The error message given by the Display representation of an error type should be lowercase without trailing punctuation, and typically concise.
```

Remove capitalization of first work in error message

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,489 @@
...
+        }
+    }
+}
+
+/// Definition of a Quad M31 in its reduced form. The internal representation
+/// is composed by the coordinates of the QM31, following a little endian ordering.
```
tdelabro, line 37:
```rust
/// Definition of a Quad M31 in its reduced form.
///
/// The internal representation is composed of the coordinates of the QM31, following a little-endian ordering.
```

The first line before the empty line will be used as the "short" documentation for the struct. While the rest will be used as a "body".

See in the `std` lib:

Image
Image
Image

Do the same everywhere.

tdelabro: maybe just `A Quad M31 in its reduced form`. 
"definition of` is a bit redundant with the fact we are reading the doc of a struct

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,489 @@
...
+#[repr(transparent)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct QM31([u64; 4]);
```
tdelabro, line 40: About the name of this struct: `QM31`.
There is two abbreviations here: `Q` for `Quad` (in the doc), I'm guessing it is for `Quadruple` and `M` for `Mersenne `.

This is a log of abbreviations. And even the doc keep mentioning those abreviation: `Quad M31 `.

I would either rename the struct to something less abreviated, or/and add a `//!`  module doc with the long version of what is defined in this module: A mersenne 31 field element represented as four u64.

tdelabro: Without this you have to already know what is contained here some other way to understand what is going on.
We should minimize prior knowledge requirements

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,489 @@
...
+
+impl QM31 {
+    /// [QM31] constant that's equal to 0.
+    pub const ZERO: Self = Self([0, 0, 0, 0]);
+
+    pub fn inner(&self) -> [u64; 4] {
+        self.0
+    }
+
+    /// Creates a [QM31] in its reduced form from four u64 coordinates.
```
tdelabro, line 50: Can you provide a link to what `coordinates` means in this context? I'm guessing it is some well known mathematical concept.

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,489 @@
...
+
+    /// Packs the [QM31] coordinates into a Felt.
+    fn pack_into_felt(&self) -> Felt {
+        let coordinates = self.0;
+
+        let bytes_part1 = (coordinates[0] as u128 + ((coordinates[1] as u128) << 36)).to_le_bytes();
+        let bytes_part2 = (coordinates[2] as u128 + ((coordinates[3] as u128) << 36)).to_le_bytes();
+        let mut result_bytes = [0u8; 32];
+        result_bytes[0..9].copy_from_slice(&bytes_part1[0..9]);
+        result_bytes[9..18].copy_from_slice(&bytes_part2[0..9]);
```
tdelabro, line 68: Can you comment what you are doing for improving readability?
I understand that qm31 is smaller that Felt so we don't need to set all bytes at the end.
But why are we using only 18 bytes? Maybe because m31::MAX only use only 144 bits. Just say it so I don't have to guess myself.

tdelabro: Also

`let bytes_part1 = (coordinates[0] as u128 + ((coordinates[1] as u128) << 36)).to_le_bytes();`

The shifting is not enough for the bits not to overlap at some points.
What algorithm are we following here? Why shift by `36`? To me (not familiar with this) it is a magic number.
Ideally I should be able to learn and understand what is the algorithm here by reading doc + comments + additional link (to wikipedia or another impementation) if needed

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,489 @@
...
+    }
+
+    /// Computes the multiplication of two [QM31] elements in reduced form.
+    pub fn mul(&self, rhs: &QM31) -> QM31 {
+        let coordinates1_u64 = self.inner();
+        let coordinates2_u64 = rhs.inner();
+        let coordinates1 = coordinates1_u64.map(u128::from);
+        let coordinates2 = coordinates2_u64.map(u128::from);
+
+        let result_coordinates = [
```
tdelabro, line 117: Can you add a link to the definition of the algorithm you are using here?
Is it a famous one or something you created yourselves?
This looks utterly random to me and any future reader of the code.

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,489 @@
...
+            coordinates1_u64[0] * coordinates2_u64[3]
+                + coordinates1_u64[1] * coordinates2_u64[2]
+                + coordinates1_u64[2] * coordinates2_u64[1]
+                + coordinates1_u64[3] * coordinates2_u64[0],
+        ];
+        Self::from_coordinates(result_coordinates)
+    }
+
+    /// Computes the inverse in the M31 field using Fermat's little theorem, i.e., returns
+    /// `v^(STWO_PRIME-2) modulo STWO_PRIME`, which is the inverse of v unless v % STWO_PRIME == 0.
```
tdelabro, line 145: split doc into

/// Short doc
///
/// Addtionnal explainations about the inner logic

crates/starknet-types-core/src/felt/qm31.rs
```diff
@@ -0,0 +1,489 @@
...
+    /// Computes the inverse of a [QM31] element in reduced form.
+    /// Returns an error if the operand is equal to zero.
+    pub fn inverse(&self) -> Result<QM31, QM31Error> {
+        if *self == Self::ZERO {
+            return Err(QM31Error::InvalidInversion);
+        }
+
+        let coordinates = self.inner();
+
+        let b2_r = (coordinates[2] * coordinates[2] + STWO_PRIME * STWO_PRIME
```
tdelabro, line 174: explain the algorithm you are using. add a link to it

tdelabro (changes requested):

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Why is it that the shitting is not enough?

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Done in: 9de3ac978c157a26d9fc93dbe8ff17f3bab1e994

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Done in: [8674230](https://github.com/starknet-io/types-rs/pull/149/commits/867423074ab980ab524e476b7d44d691e182572b)

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Changed a little the docs in this struct: [8674230](https://github.com/starknet-io/types-rs/pull/149/commits/867423074ab980ab524e476b7d44d691e182572b). Let me know if you prefer changing the name.

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Extended the comment here: [8674230](https://github.com/starknet-io/types-rs/pull/149/commits/867423074ab980ab524e476b7d44d691e182572b)

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Done a little correction, better to check this link: https://github.com/lambdaclass/types-rs/blob/a65f3668cae00e0dabb60a8a8725a530a2993acd/crates/starknet-types-core/src/qm31/mod.rs#L42. I previously mentioned little endian in the comment by accident when it should be big endian.

crates/starknet-types-core/src/qm31/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+            QM31Error::FeltTooBig(felt) => write!(
+                f,
+                "number used as QM31 since it's more than 144 bits long: {felt}"
+            ),
+            QM31Error::InvalidInversion => write!(f, "attempt to invert a qm31 equal to zero"),
+        }
+    }
+}
+
+/// Definition of a Quadruple Merseene 31.
```
tdelabro, line 36: mersenne

tdelabro: Aa Quadruple Merseene 31

crates/starknet-types-core/src/qm31/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    /// [QM31] constant that's equal to 0.
+    pub const ZERO: Self = Self([0, 0, 0, 0]);
+
+    pub fn inner(&self) -> [u64; 4] {
+        self.0
+    }
+
+    /// Creates a [QM31] in its reduced form from four u64 coordinates.
+    /// 
+    /// A coordinate refers to a value in the M31 field.
```
tdelabro, line 54: I still struggle to understand what is the mathematical relationship between those coordinates and the number.

Is it: `felt_value = c₀ + c₁ × 2³⁶ + c₂ × 2⁷² + c₃ × 2¹⁰⁸` ?
Claude deduced this by looking at `pack_into_felt`

If so, make it explicit

tdelabro (changes requested):

crates/starknet-types-core/src/qm31/mod.rs
FrancoGiachetta: Yes a felt should be equal to that. Do you want me to add that in `pack_into_felt` doc?

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Added some docs from this here: [dde739b](https://github.com/starknet-io/types-rs/pull/149/commits/dde739b945f3b9c3ffd391559be5ed46aced148d)

crates/starknet-types-core/src/felt/qm31.rs
FrancoGiachetta: Added some docs from this here: [dde739b](https://github.com/starknet-io/types-rs/pull/149/commits/dde739b945f3b9c3ffd391559be5ed46aced148d)

crates/starknet-types-core/src/qm31/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+    }
+}
+
+/// Definition of a Quadruple Mersenne 31.
+///
+/// The internal representation is composed of 4 limbs, following a big-endian ordering.
+/// Each of this limbs can be represented by 36 bits.
+#[repr(transparent)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct QM31([u64; 4]);
```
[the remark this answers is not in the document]
FrancoGiachetta: Change the representation to u32 coordinates [020c0dc](https://github.com/starknet-io/types-rs/pull/149/commits/020c0dcc7501c07f95b2076b2afbc5bc1b0d41d8).

crates/starknet-types-core/src/qm31/mod.rs
```diff
@@ -0,0 +1,543 @@
...
+/// Each of this limbs can be represented by 31 bits.
+#[repr(transparent)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct QM31([u32; 4]);
+
+impl QM31 {
+    /// [QM31] constant that's equal to 0.
+    pub const ZERO: Self = Self([0, 0, 0, 0]);
+
+    pub fn inner(&self) -> [u32; 4] {
```
gabrielbosio, line 49:
```suggestion
    pub fn to_coordinates(&self) -> [u32; 4] {
```

crates/starknet-types-core/src/qm31/mod.rs
```diff
@@ -0,0 +1,543 @@
...
+        self.0
+    }
+
+    /// Utility method to return the coordinates of a [QM31] as u64 values.
+    ///
+    /// This method is convinient for performing multications and inversions,
+    /// in which operating with the coordinates could result in overflows if
+    /// the 32 bit representation was used.
+    fn inner_u64(&self) -> [u64; 4] {
```
gabrielbosio, line 58:
```suggestion
    fn to_coordinates_u64(&self) -> [u64; 4] {
```

gabrielbosio (changes requested):

crates/starknet-types-core/src/qm31/mod.rs
tdelabro: convinient -> convenient

crates/starknet-types-core/src/qm31/mod.rs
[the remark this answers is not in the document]
FrancoGiachetta: I mess up with the docs in cairo VM. This firstly was implemented in the cairo VM and since it only operates with felts, a qm31 was always being packed into a Felt. For efficiency reasons (check this https://github.com/lambdaclass/cairo-vm/pull/1944#discussion_r1953261265), related to the multiplication operation, it's better to represent every M31 with 36 bits instead of 31 (as it should). This mens that, only in the felt representation, a QM31  is 144 bits long.

This implementation treats every M31 as 31 bit value. So only when packing a QM31 into a felt a M31 be represented with 36 bits now. This is why we still check that, when unpacking from a Felt, the value is no longer than 144 bits.

FrancoGiachetta:
Hi @tdelabro! Added the requested top level docs.

crates/starknet-types-core/src/qm31/mod.rs
```diff
@@ -1,3 +1,20 @@
+//! A Cairo-like QM31 type.
+//!
+//! The QM31 type represents a Degree-4 extension the Mersenne 31 Field. This extension can be
+//! represented as two components of the Complex extension the Mersenne 31 Field as follows ((a, b), (c, d)), where
+//! a, b, c and d represent a value from the Mersenne 31 Field, denotated as M31.
+//! If anly a M31 was used by the verifier, then a 31 bit value wouldn't be enough to provide security
```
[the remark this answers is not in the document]
FrancoGiachetta: It refers to the value a named. This is to avoid repeating "a value from the Mersenne 31 field". The first typo is already fixed in this commit: [b65e34b](https://github.com/starknet-io/types-rs/pull/149/commits/b65e34b7b6298ef992f5f4ef8c81518dc7aa8267)

crates/starknet-types-core/src/qm31/mod.rs
```diff
@@ -0,0 +1,561 @@
+//! A Cairo-like QM31 type.
+//!
+//! The QM31 type represents a Degree-4 extension in the Mersenne 31 field. This extension can be
+//! represented as two components of the Complex extension the Mersenne 31 field as follows ((a, b), (c, d)), where
```
gabrielbosio, line 4: > of the Complex extension the Mersenne 31 field

It doesn't seem very clear, maybe a word is missing here?

crates/starknet-types-core/src/qm31/mod.rs
```diff
@@ -0,0 +1,561 @@
...
+//! a, b, c and d represent a value from the Mersenne 31 field, denotated as M31.
+//! If only a M31 was used by the verifier, then a 31 bit value wouldn't be enough to provide security
+//! to the verification. A QM31 not only provides an efficient arithmetic field, since it is composed of four M31 values, but
+//! also allows for a more secure level of verification as it offers a 124 bit value. By using this extension,
+//! the verifier is able to generate challeges with a proper level of randomness, ensuring the security of the protocol.
```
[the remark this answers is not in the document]
FrancoGiachetta: A verifier in general. A 31 bit value is not enough to provided security since the range of random values is too short and vulnerable to brute-force attacks

crates/starknet-types-core/src/qm31/mod.rs
```diff
@@ -0,0 +1,561 @@
...
+//!
+//! While the Cairo language's representation of QM31 consists of four `BoundedInt`s
+//! simulating a 31 bit value, this implementation uses four 32 bit values.
+//!
+//! The conversion to a Felt is done by using the four elements of the struct, refered to as coordinates, as bytes and then 
+//! parsed as a little endian value. For a more efficient Felt representation, each coordinate is stored as a 36 bit. Hence,
```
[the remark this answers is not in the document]
FrancoGiachetta: Done here: [6bad58f](https://github.com/starknet-io/types-rs/pull/149/commits/6bad58f46ed8f08ac2f9ad8ac34de539155389bf)

crates/starknet-types-core/src/qm31/mod.rs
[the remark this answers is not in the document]
FrancoGiachetta: fixed typo here: [6bad58f](https://github.com/starknet-io/types-rs/pull/149/commits/6bad58f46ed8f08ac2f9ad8ac34de539155389bf). Also changed `the verifier` for `a verifier` as the comment aims to a verifier in general.

PR #152, opened 2025-08-22
Add QM31 implementation (Lambdaworks)

JulianGCalderon:
This PR depends on https://github.com/starknet-io/types-rs/pull/151. As the base of this PR is main, the diff includes the changes of the #151.

This PR implements the QM31 newtype, as a wrapper of Lambdawork's type.

As the actual implementation is from Lambdaworks, I only implemented the `pack_into_felt` and `unpack_from_felt` methods. Should I also implement wrappers for common operations? (i.e. mul, add, inv, sub).

# Pull Request type

- Feature

We have no implementation of QM31

## What is the new behavior?

Implements QM31 as a newtype of Lambdawork's type.

No


crates/starknet-types-core/src/qm31/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+
+    fn is_zero(&self) -> bool {
+        self == &Self::zero()
+    }
+}
+impl One for QM31 {
+    fn one() -> Self {
+        Self(FieldElement::<Degree4ExtensionField>::one())
+    }
+}
```
tdelabro, line 142: Implementation of traits comming from `num-traits` crate should be behind a feature flag.
Check the way we did it for Felt in `crates/starknet-types-core/src/felt/num_traits_impl.rs`

tdelabro (changes requested):

0xLucqs (approved):

PR #156, opened 2025-09-02
fix: MSB for big numbers

famouswizard:
the marker for big numbers was applied at the wrong place, which could mess up the most significant bit.
now it’s always set on the first element, so 256-bit numbers match Cairo’s little-endian format.
removed the unused `start` variable and updated `test_encode_felts_to_u32s_big` to check the marker.


tdelabro:
hey @famouswizard  this is great!
Can you provide a test that prove that there was a bug. I confess that I struggle to get what is going on haha

AvivYossef-starkware (changes requested):
@famouswizard  plz TAL the cairo implementation.

PR #171, opened 2025-10-06
feat: add ChainId type

tdelabro:
Add the type `ChainId`

## What is the new behavior?

The `ChainId` type allows for representing the two starknet chains, SN_MAIN and SN_SEPOLIA.
It offers type conversion from and into Felt, ShortString, String.

A new Rust feature `devnet` allows adding a custom local devnet with any valid `ShortString` as name (eg. `SN_DEVNET`).

No


crates/starknet-types-core/src/chain_id/alloc_impls.rs
```diff
@@ -0,0 +1,191 @@
...
+#[cfg(feature = "devnet")]
+impl From<ShortString> for ChainId {
+    fn from(value: ShortString) -> Self {
+        if value.as_ref() == SN_MAIN_STR {
+            ChainId::Mainnet
+        } else if value.as_ref() == SN_SEPOLIA_STR {
+            ChainId::Sepolia
+        } else {
+            ChainId::Devnet(value)
+        }
```
gabrielbosio:
```suggestion
        match value.as_ref() {
            SN_MAIN_STR => ChainId::Mainnet,
            SN_SEPOLIA_STR => ChainId::Sepolia,
            _ => ChainId::Devnet(value),
        }
```

Also, why does conversion from `ChainId` to `ShortString` is possible but not the other way around if feature `devnet` is disabled?

crates/starknet-types-core/src/chain_id/alloc_impls.rs
```diff
@@ -0,0 +1,191 @@
...
+
+#[cfg(not(feature = "devnet"))]
+impl From<ChainId> for &str {
+    fn from(value: ChainId) -> Self {
+        match value {
+            ChainId::Mainnet => SN_MAIN_STR,
+            ChainId::Sepolia => SN_SEPOLIA_STR,
+        }
+    }
+}
```
gabrielbosio: Why conversion from `ChainId` to string is allowed only if `devnet` feature is disabled?

crates/starknet-types-core/src/short_string/mod.rs
```diff
@@ -42,7 +50,45 @@ impl From<ShortString> for Felt {
...
+        }
+        let first_non_zero_byte = match bytes.iter().position(|&v| v != 0) {
+            Some(i) => i,
+            None => return Ok(ShortString(String::new())),
+        };
+        if !bytes[first_non_zero_byte..].is_ascii() {
+            return Err(TryShortStringFromFeltError::NonAscii);
+        }
+
+        let s = unsafe { str::from_utf8_unchecked(&bytes[first_non_zero_byte..]) };
```
gabrielbosio: I would add a comment explaining why is it safe to call `from_utf8_unchecked` here (It is because of the condition in lines 81-83, right?)

crates/starknet-types-core/src/chain_id/alloc_impls.rs
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        #[cfg(feature = "devnet")]
+        match ShortString::try_from(value) {
+            Ok(ss) => Ok(ChainId::Devnet(ss)),
+            Err(e) => Err(TryChainIdFromStringError(e)),
+        }
+
+        #[cfg(not(feature = "devnet"))]
+        Err(TryChainIdFromStringError(value))
```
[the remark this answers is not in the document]
tdelabro: I tried to use match more accross the board

crates/starknet-types-core/src/chain_id/alloc_impls.rs
tdelabro: Yep.
with devnet:
- any shortstring is a valid devnet name (we don't enforce it has to start with the `SN_` prefix, maybe we should, but I don't have strong opinion about that)
- TryFrom is automatically implemented as infaiilbe and will resolve to the From impl

without devnet:
- only two known values are valid `SN_MAIN_STR` and `SN_SEPOLIA_STR`, all other will result in an error
- therefor `From` is not implemented, only TryFrom

tdelabro:
@maciejka @0xLucqs @dorimedini-starkware @AvivYossef-starkware need one last review here

crates/starknet-types-core/src/chain_id/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+#[cfg(not(feature = "devnet"))]
+pub struct TryChainIdFormFeltError(Felt);
+
+#[derive(Debug, Clone, Copy)]
+#[cfg(feature = "devnet")]
+pub struct TryChainIdFormFeltError(crate::short_string::TryShortStringFromFeltError, Felt);
```
dorimedini-starkware: typo

crates/starknet-types-core/src/chain_id/alloc_impls.rs
```diff
@@ -0,0 +1,198 @@
...
+use crate::short_string;
+use crate::short_string::ShortString;
+
+use super::{ChainId, SN_MAIN_STR, SN_SEPOLIA_STR};
+
+impl From<ChainId> for ShortString {
+    fn from(value: ChainId) -> Self {
+        match value {
+            ChainId::Mainnet => short_string!("SN_MAIN"),
+            ChainId::Sepolia => short_string!("SN_SEPOLIA"),
```
dorimedini-starkware: you define these strings as constants in another module, please reuse them

crates/starknet-types-core/src/short_string/mod.rs
```diff
@@ -116,10 +169,88 @@ impl TryFrom<&str> for ShortString {
...
+///
+/// let ss = short_string!("Hello, Cairo!");
+/// assert_eq!(ss.to_string(), "Hello, Cairo!");
+///
+/// // This would fail to compile:
+/// // let ss = short_string!("This string is way too long for a Cairo short string");
+/// ```
+#[macro_export]
+macro_rules! short_string {
+    ($s:literal) => {{
```
dorimedini-starkware: why must this be a literal? IIUC the macro allows any const &str

crates/starknet-types-core/src/chain_id/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+#[cfg(feature = "alloc")]
+pub use alloc_impls::*;
+
+#[derive(Debug, Clone)]
+pub enum ChainId {
+    Mainnet,
+    Sepolia,
+    #[cfg(feature = "devnet")]
+    Devnet(ShortString),
+}
```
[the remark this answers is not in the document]
tdelabro: https://docs.starknet.io/learn/cheatsheets/chain-info
No mention of it in the official starknet documentation.

https://github.com/starkware-libs/starknet-api/blob/91fcc564b125c16ada9d956dcbfa2628415ca011/src/core.rs#L26
The code itself doesn't explain what it should be used for.

Can you tell me which network it describes, when it should be used, and can you ping someone in charge of Starknet documentation to add it to the official website?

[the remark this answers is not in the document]
dorimedini-starkware: I may need more context here: there is a `ChainId` enum on the sequencer repo, is this PR intended to be a replacement?
We have an integration system which is intended for upgrade preparations, not really useful for devs and users in general

